PR #4, opened 2015-11-06
DLL Build Project for Visual Studio

uraymeiviar:
Added Visual Studio 2015 project, and also added new project configuration to build DLL instead of static library


.gitignore
```diff
@@ -78,3 +78,5 @@ silk/float/x64
 src/Debug
 src/Release
 src/x64
+*.opensdf
+win32/
```
rillian: I don't think you should ignore win32 here? It contains tracked files.

MisterZeus: rillian, you're right. Still, this pull could be applied as it's got some good stuff, and the .git ignore line removed after?

rillian:
Ok, I've pushed the change and then reverted the .gitignore change. Thanks for contributing!

PR #9, opened 2016-05-11
Fix VS2015 dynamic builds by adding DLL_EXPORT to CFLAGS

wiiaboo:
Needed after xiph/opus@6e96173.


rillian:
thanks! merged in 5ff11c02dabe4c857ab39f38b557782999d23224.

PR #10, opened 2016-05-13
DLL fixes and fopen/fopen_s error suppression

MisterZeus:
Tested all 4 configurations, including the 2 new DLL ones, across Win32 and x64 platforms. All 9 projects now build fine across all 8 combinations!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -75,25 +75,3 @@ silk/float/x64
 src/Debug
 src/Release
 src/x64
-*.opensdf
-
-# Visual Studio ignores
-[Dd]ebug/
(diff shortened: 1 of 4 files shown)


wiiaboo:
As a reminder from IRC:
- There are still other `$(PlatformTarget)` left unchanged in opus.vcxproj, is this intended?
- Remake the first commit to only have the fixes/changes to the two .vcxproj files, merge the .gitignore fixes and separation to its own commit and remove version.mk changes from both commits.

MisterZeus:
Thanks! Now fixed, hopefully.

PR #14, opened 2016-07-08
vs2015: supress a few warnings

wiiaboo:
SDLCheck is a superset of BufferSecurityCheck and is off by default.
If it's set, it complains that it's overriden by BufferSecurityCheck.

Warning 4996 is already ignored in other binaries (fopen being deprecated and
suggesting fopen_s).

NoExtensions isn't a valid value for EnableEnhancedInstructionSet in x64 builds.

### DIFF
--- a/win32/VS2015/celt.vcxproj
+++ b/win32/VS2015/celt.vcxproj
@@ -324,7 +324,6 @@
       <ControlFlowGuard>Guard</ControlFlowGuard>
       <FunctionLevelLinking>false</FunctionLevelLinking>
       <EnableParallelCodeGeneration>false</EnableParallelCodeGeneration>
-      <EnableEnhancedInstructionSet>NoExtensions</EnableEnhancedInstructionSet>
       <FloatingPointExceptions>true</FloatingPointExceptions>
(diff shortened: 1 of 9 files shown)


Chocobo1:
>vs2015/opus: Use precise floating-point instead of fast

would it be better to specify fp:precise for that function only? since other functions pass the tests with fp:fast. e.g.
```c
#pragma float_control(precise, on, push)
OPUS_EXPORT void opus_pcm_soft_clip(float *_x, int N, int C, float *declip_mem) {...}
#pragma float_control(pop)
```

BTW, at https://github.com/xiph/opus/blob/master/tests/test_opus_decode.c#L403
```c
      for (j=0;j<(1024/i)*i;j++)
      {
        if(x[i]>1.f)test_failed();   // should this be x[i] => x[j] ??
        if(x[i]<-1.f)test_failed();  // here too ??
      }
```

wiiaboo:
fp:fast was only used in silk_common before the VS2015 transition. All other projects used fp:precise.
I only tested building the projects, not running the test binaries back then.

If it's a bug in the test (reproduceable with GCC and `-ffast-math` too) and it's fixed, I can drop the second commit.

PR #15, opened 2016-07-11
src/opus: Force precise floating-point for opus_pcm_soft_clip (vs2015)

wiiaboo:
Fixes failed test in test_opus_decode when using fp:fast.
Suggested in #14

This test also fails when compiled with GCC 5.4 with -ffast-math.


wiiaboo:
Either that or fixing the bug with the function and lower precision floats.

PR #17, opened 2016-07-27
VS2015 Merge opus dependencies into its project

wiiaboo:


rillian:
Merged.

wiiaboo:
To do: maybe use an easier host for the artifacts. The artifacts need the appveyor token to be downloaded programatically using appveyor's REST api and even then I couldn't figure out how within the build (worked fine locally using PowerShell).

PR #25, opened 2017-01-10
VS2015/opus: remove dead options for linker in static configurations

wiiaboo:


wiiaboo:
Included in https://github.com/xiph/opus/pull/26.

PR #26, opened 2017-01-11
[RFC] VS2015: Rework all common options into a props file

wiiaboo:
After a few hours of reading MS documentations and regex on the project files, I've put together most/all common options in the .vcxproj files into a single common.props that's included in every configuration.

I'd very much appreciate someone more experienced with MSVC tested this and/or mentioned any misgivings about doing this.

At first look, this makes the VS files considerably easier to understand and gives a big savings in repeated code, but other than batch building every project I can't be sure I missed anything.

Before merging I'll squash the two commits, this is not for merge as-is.


wiiaboo:
Note: this common.props file can be reused with the other projects to also deduplicate unneeded code and remove maybe wrong/useless options.

PR #27, opened 2017-01-21
add more coverage to travis tests

stephengroat:


rillian:
I still think this isn't worth the cpu time, but I guess it's not the limiting factor in patch review, so r=me for more testing. Merged as 4176a0c86e492bd8cf8e88e209ef5c9be3678c9f.

PR #34, opened 2017-02-27
Silence last warnings

wiiaboo:


jmvalin:
I can't verify that the project file is correct, but I agree with the idea of fixing these warnings from the build system.

PR #38, opened 2017-04-10
Avoid warning when __STDC_VERSION__ is not defined

clopez:
This fixes a warning when building with GCC:
```

[....]  include/opus_types.h:37:39: error: "__STDC_VERSION__" is not defined [-Werror=undef]
 #if (defined(__STDC__) && __STDC__ && __STDC_VERSION__ >= 199901L) || (defined(__GNUC__) && (defined(_STDINT_H) || defined(_STDINT_H_)) || defined (HAVE_STDINT_H))
                                       ^

```

### DIFF
--- a/include/opus_types.h
+++ b/include/opus_types.h
@@ -34,7 +34,7 @@
 #define OPUS_TYPES_H
 
 /* Use the real stdint.h if it's there (taken from Paul Hsieh's pstdint.h) */
-#if (defined(__STDC__) && __STDC__ && __STDC_VERSION__ >= 199901L) || (defined(__GNUC__) && (defined(_STDINT_H) || defined(_STDINT_H_)) || defined (HAVE_STDINT_H))
+#if (defined(__STDC__) && __STDC__ && defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L) || (defined(__GNUC__) && (defined(_STDINT_H) || defined(_STDINT_H_)) || defined (HAVE_STDINT_H))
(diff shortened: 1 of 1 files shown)


rillian:
Hi! Thanks for the report. I've not seen this issue before. Can you tell me which version of gcc this happens with?

According to the [manual](https://gcc.gnu.org/onlinedocs/cpp/Standard-Predefined-Macros.html#Standard-Predefined-Macros) this macro is defined unless one passes `-traditional-cpp` or are compiling as C++ without an `extern "C"` block (which will cause other problems). Could you be doing either of those things?

rillian:
I guess `-traditional-cpp` also disables `__STDC__` so it can't be that.

clopez:
At least GCC 4.9 (GCC 4.9.2 as shipped in Debian stable) defines `__STDC__` but not `STDC_VERSION`. 

```
$ g++-4.9 -dM -E - < /dev/null | grep STDC
#define __STDC_HOSTED__ 1
#define __STDC_IEC_559__ 1
#define __STDC_ISO_10646__ 201103L
#define __STDC_NO_THREADS__ 1
#define _STDC_PREDEF_H 1
#define __STDC_IEC_559_COMPLEX__ 1
#define __STDC__ 1
```

rillian:
So it doesn't. Thanks for clarifying. I'll add this to the commit message and push.

rillian:
Merged as 1f8cea2d13b21e55518e55c15ec6bd9510164696. Thanks for the fix!

PR #41, opened 2017-05-10
Suggest basic build dependencies in the README.

rillian:
Make it easier for users unfamiliar with C applications
to installed the necessary build dependencies.


gcp:
I'd add indentation to the distro/platform specific commands, makes it easier to find the one applicable to you, and currently it's a bit hard to read.

PR #46, opened 2017-06-13
Fix a duplicate symbol name.

rillian:
Both versions of silk/tables_NLSF_CB had symbols
named silk_NLSF_CB1_Wght_Q9 which causes problems
if both files are build as part of the same
compilation unit. Rename the wideband version
to include a 'WB' string to avoid the conflict.


rillian:
Merged in 34d821ceae87bd8baf8eedbd8037ef3dc05455a5.

PR #57, opened 2017-08-10
Force crlf line endings for VS2015 project files

pps83:
this ensures that regardless of git settings VS project files have correct line endings. This make VS to use correct version of visual studio.


rillian:
Merged in 1f38af47b5eec7d8adfc5481b0d1be67bc6ad96e. Thanks!

PR #59, opened 2017-08-13
Visual Studio 2017 solution

Zuzu-Typ:
This pull request adds a Visual Studio 2017 solution.


rillian:
Thanks for contributing this. If VS2017 can open VS2015 project files, isn't it easier to just have one set and let people open it with whichever version of Visual Studio they're using?

Zuzu-Typ:
I guess that makes sense, that way there is no issue with having to search in VS2017 / VS2015 subfolders. I will quickly adapt the other pull requests.
This one can be closed

rillian:
ok, thanks!

PR #64, opened 2017-09-11
Fix typo with NE10 prefix

akochetkov:
Prefix passed using '--with-NE10=PFX' not used as NE10 include path.

Signed-off-by: Alexander Kochetkov <al.kochet@gmail.com>


rillian:
Thanks for the fix! I've reviewed and pushed it as ec29ffd94407864e565956e1fa0f834c43502508.

PR #79, opened 2018-02-27
Proper detection of _mm_cvtss_si32 for MS compiler

pps83:
_mm_cvtss_si32 is available starting from SSE


jmvalin:
I don't think the defined(_M_IX86_FP) part is needed since _M_IX86_FP will automatically expand to 0 if undefined. Also why the explicit AVX test? Are there cases when it triggers without the other ones being defined?

pps83:
I agree, kind of strange, but I took the check as-is from win32/config

jmvalin:
Merged the first part (_mm_cvtss_si32) for 1.3. I'll handle the rest in a future release.

jmvalin:
Rest of the patch merged in master

PR #80, opened 2018-02-27
Remove requirement to add fixed and float dirs to include path

pps83:
I didn't try to build other platforms, works for me on mac/windows.

In case if somebody simply includes opus src code this change makes life easier.


jmvalin:
I'm not sure I understand the purpose of this patch

pps83:
It's not like it fixes a bug, more style related cleanup. opus has a few dirs with source code and pretty much all of them are added to include paths. If you don't think it's needed, just close PR

PR #83, opened 2018-03-26
Rename celt's "compute_allocation" to "compute_pulse_allocation" to avoid symbol conflicts

maxw:
I use opus as a static library along with another static library that defines the symbol "compute_allocation". Ideally I'd rather just prefix all celt functions with "celt_", but I figure this was the most minimal change I could make.

### DIFF
--- a/celt/celt_decoder.c
+++ b/celt/celt_decoder.c
@@ -1026,7 +1026,7 @@ int celt_decode_with_ec(CELTDecoder * OPUS_RESTRICT st, const unsigned char *dat
    ALLOC(pulses, nbEBands, int);
    ALLOC(fine_priority, nbEBands, int);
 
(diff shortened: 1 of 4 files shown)


maxw:
For some reason I assumed this would open a pull request on my github fork. Ideally these four commits should be squashed into one before merging.

jmvalin:
Done (with clt_ prefix).

maxw:
Awesome! Where has the fix been made already? I don't see any changes to master on github. Should I be looking somewhere else?

Max

rillian:
I think Jean-Marc meant he fixed it his own way, but the github mirror took some time to propagate. The change is in aebc9046f9902e0c3f44d0d866261294592ea693 above. Sorry for the confusion!

PR #84, opened 2018-03-27
Avoid DTX when Silk sees activity

gustafullberg:
This PR fixes an issue with periodic clicks or noise-bursts during DTX that we have seen in Chrome, AppRTC and other WebRTC clients.

Sometimes, Opus (in silk-only or hybrid mode) decides to go into DTX even though Silk considers the signal to contain activity. This is because the `decide_dtx_mode` has a different VAD than Silk.

In DTX, Opus sends an encoded packet every 420 ms. This is in order to give an updated estimation of the background noise. The decoder uses comfort noise (CNG) to produce output during the DTX period.

However, when the silk layer of the "update packets" that arrive every 420 ms contain activity (type `TYPE_UNVOICED` or `TYPE_VOICED` instead of `TYPE_NO_VOICE_ACTIVITY`) the decoder will instead use packet loss concealment (PLC) to produce output instead of CNG. The PLC is then faded out as CNG ramps up.

In cases when several update packets contain activity, the switching back and forth between PLC and CNG will cause an audible click or noise-burst every 420 ms.

This PR fixes the issue by preventing Opus from entering DTX in hybrid or silk-only mode whenever silk considers the signal to be active. Celt-only mode is unaffected.

### DIFF
--- a/src/opus_encoder.c
+++ b/src/opus_encoder.c
@@ -2129,6 +2129,12 @@ opus_int32 opus_encode_native(OpusEncoder *st, const opus_val16 *pcm, int frame_
 #ifndef DISABLE_FLOAT_API
     if (st->use_dtx && (analysis_info.valid || is_silence))
     {
+       /* Mark the frame active if Silk considers it active */
+       if(st->mode != MODE_CELT_ONLY && st->silk_mode.signalType != TYPE_NO_VOICE_ACTIVITY)
+       {
+          analysis_info.activity_probability = 1.0f;
+       }
+
        if (decide_dtx_mode(analysis_info.activity_probability, &st->nb_no_activity_frames,
              st->peak_signal_energy, pcm, frame_size, st->channels, is_silence, st->arch))
        {


gustafullberg:
Spectrograms illustrating the issue:

dtx_clicks

jmvalin:
Unless I missed something (which is quite possible), it seems like the opposite approach would make more sense. By that I mean forcing SILK to consider the signal as inactive when DTX got enabled by Opus. This would have two advantages:
1) AFAIK the Opus VAD is more reliable than the SILK one
2) It would ensure the DTX decision is the same in all modes, whereas your patch would make SILK use DTX less than CELT (since it can turn DTX off)
Does that make sense or are there other reasons you're making the change the other way?

gustafullberg:
That's a good point. I think that could be the right way of solving this.

I created #87 that solves the problem in the way you mention. If your prefer that solution, that PR could replace this one.

The new PR touches a few more lines but is still pretty clean.
This PR is a bit more conservative when it comes to DTX in non-stationary noise.

jmvalin:
merged the improved patch

PR #87, opened 2018-04-10
Silk makes use of Opus VAD

gustafullberg:
This pull request is an alternative solution to the problem described in PR #84 . It was created in response to Jean-Marc Valin's comment.

The PR makes Silk use the Opus VAD when available (float API enabled, complexity level high enough). When the Opus vad is not used the Silk VAD will be used instead.

Using the same VAD for Silk and Opus solves the issue with the "click sounds", described in PR #84, that are caused by PLC usage in the DTX region.

There is also some reduction of complexity due to running one VAD instead of two.
Behavior of complexity modes where the Opus VAD is not run is unchanged.


gustafullberg:
Updated the PR to fix some issues.

PR #97, opened 2018-09-10
RC finds

ProDigit:
Hi,

I find significant sound quality improvement of Stereo audio of 64kbps or below, and mono of 20kbps or below, when setting the framesize to 40, vs the stock 20.
This is on the RC (1.3). 
Several low bitrate files sound much better.

The file size savings are substantial at about 6%!
And while the old codec sounded ok for portable versions at 48kbps, the newer RC sounds massively better adding only 3 bytes to the 48kbps to 51kbps, while maintaining round about the same filesize as the older codec at 48kbps!

11kbps mono or 22kbps stereo can be gotten with ~32kHz samples, as opposed to 22kHz samples with framerate 20. 

I see only benefits, from both audio quality improvements, and file size reductions point of view.

I think on the RC, lower than 64kbps, the system default should be --Framesize 40!


jmvalin:
What exactly is that pull request that includes all the commits in the past 2+ years??? Also, if you're seeing major differences between 20 ms frames and 40 ms frames, then either something's wrong with your experiment or you're measuring just at the threshold of an encoder decision (and the frame size changes the bitrate for that threshold by a few bits/second)

ProDigit:
I have no idea how to even use this site.
But the chat is offline. Too much spam was mentioned, so I came here to post.
Yes,Think that's what happens. The RC is a great improvement over the standard version, which pulls the bitrates to the threshold trying to get the highest sound quality, and it exceeds my expectations.
However, at these settings, frame size 40 makes a big difference.
20 or 60 sound just plain worse. Especially noticeable at lower bitrates.

jmvalin:
When you ask for 40-ms packets, Opus just codes 20-ms frames and packs then together in a packet, so there's no reason why it should sound very different. Can you upload an obvious example? Please include the input file, the exact command line (and version), as well as the (compressed) output files.

jmvalin:
Look, I don't know where you got that 6% figure, because the differences I get are around 0.3%. So you can either provide exact examples of input/output files with the exact versions and steps to reproduce, or otherwise it's not worth discussing this.

PR #99, opened 2018-10-22
Trivial sprintf to snprintf conversion.

icepic:
Some linkers warn about unbounded sprintf calls, like the OpenBSD linker.

Since snprintf() is already used in other places in the same file, I assume availability of snprintf() isn't an issue, and the size of the buffer is knowable at that point so, trivial conversion, and slightly less warnings while compiling on openbsd.

Tested on macosx x86_64 (took a minute or so), and OpenBSD/mips64 (took the whole weekend) with all 14 tests passing ok on "make check".


jmvalin:
Merged in master, thanks.

PR #100, opened 2018-11-01
Add CMake support

xnorpx:
Yet another proposal for cmake

Update

Included:
- Parse make files for sources
- Some tests that passes
- Tested on Windows 64bit debug, Linux 64 bit and Mac High Sierra
- Version parsing from git
- SSE support
- Install

Known things missing:
- Probably some build flags for Linux?
- iOS need toolchain file?
- Fixes for W4 warnings  on Windows :)

How to run:

1. cd opus
2  mkdir build
3. cd build
4. cmake ./../ -DBUILD_TESTING=1
5. make
6. ctest
7. sudo make install

-- Install configuration: "Debug"
  -- Installing: C:/Program Files (x86)/opus/lib/opus.lib
  -- Installing: C:/Program Files (x86)/opus/lib/opus.dll
  -- Installing: C:/Program Files (x86)/opus/lib/opus_static.lib
  -- Up-to-date: C:/Program Files (x86)/opus/include/opus
  -- Up-to-date: C:/Program Files (x86)/opus/include/opus/opus.h
  -- Up-to-date: C:/Program Files (x86)/opus/include/opus/opus_custom.h
  -- Up-to-date: C:/Program Files (x86)/opus/include/opus/opus_defines.h
  -- Up-to-date: C:/Program Files (x86)/opus/include/opus/opus_multistream.h
  -- Up-to-date: C:/Program Files (x86)/opus/include/opus/opus_projection.h
  -- Up-to-date: C:/Program Files (x86)/opus/include/opus/opus_types.h

Install the project...
-- Install configuration: ""
-- Installing: /usr/local/lib/libopus.so
-- Installing: /usr/local/lib/libopus.a
-- Up-to-date: /usr/local/include/opus
-- Up-to-date: /usr/local/include/opus/opus.h
-- Up-to-date: /usr/local/include/opus/opus_custom.h
-- Up-to-date: /usr/local/include/opus/opus_defines.h
-- Up-to-date: /usr/local/include/opus/opus_multistream.h
-- Up-to-date: /usr/local/include/opus/opus_projection.h
-- Up-to-date: /usr/local/include/opus/opus_types.h


xnorpx:
ping @jmvalin @berkus

berkus:
I don't find clutches like this reasonable, but if @jmvalin likes it - good for him.

As far as I understand this will break change tracking for cmake, making it either always regenerate the build system (good luck if you use MSVC or Xcode - msvc just crashes when studio project is re-generated from underneath it on rebuild) or miss some changes in the file lists.

Even though file lists are not gonna change often, the former problem of MSVC crashing is much more severe.

xnorpx:
Agree, not ideal but probably worth the tradeoff in this case.

I will try the scenario you describe in MSVC and if it is indeed crashing that should be fixed by that team.

evpobr:
@xnorpx , look [here](https://github.com/evpobr/opus/blob/cmake/CMakeLists.txt), i've implemented some options and configure tests.

xnorpx:
@evpobr Thanks! once I get a initial like from @jmvalin I will add the options.

jmvalin:
So this at least looks sane from my point of view at least (not knowing anything about cmake). I had a few questions though?
- Why does the patch rely on Makefile.unix?
- Is there a way to avoid explicitly naming all the SOURCES_SSE* sets of files?
- Any way to avoid having to name all the test_* programs (easy to forget to add one)?
None of the issues I'm raising above is fatal, but it would reduce the maintenance burden (and likelihood of messing up) if they can be addressed. 

In any case, this is looking like something I should be able to merge (assuming it works well and all, which I haven't tested yet).

jmvalin:
> As far as I understand this will break change tracking for cmake, making it either always regenerate the build system (good luck if you use MSVC or Xcode - msvc just crashes when studio project is re-generated from underneath it on rebuild) or miss some changes in the file lists.

@berkus I assume that by "break change tracking" you mean that things will break *if* someone changes the file list, as opposed to never properly rebuilding only the files that change, right? If that's the case, then I guess it's acceptable. If incremental builds is completely broken, then it's probably not acceptable.

opus_sources.cmake
```diff
@@ -0,0 +1,95 @@
...
+get_opus_sources(CELT_SOURCES_SSE4_1 celt_sources.mk celt_sources_sse4_1)
+get_opus_sources(CELT_SOURCES_ARM celt_sources.mk celt_sources_arm)
+get_opus_sources(CELT_SOURCES_ARM_ASM celt_sources.mk celt_sources_arm_asm)
+get_opus_sources(CELT_AM_SOURCES_ARM_ASM celt_sources.mk
+                 celt_am_sources_arm_asm)
+get_opus_sources(CELT_SOURCES_ARM_NEON_INTR celt_sources.mk
+                 celt_sources_arm_neon_intr)
+get_opus_sources(CELT_SOURCES_ARM_NE10 celt_sources.mk celt_sources_arm_ne10)
+
+get_opus_sources(OPUSDEMO_SRCS_C Makefile.unix opus_demo)
```
xnorpx: copy paste error: remove duplicated entries from .unix only use from Makefile.am

xnorpx:
"Why does the patch rely on Makefile.unix?"
Missed to remove it, will clean up.

"Is there a way to avoid explicitly naming all the SOURCES_SSE* sets of files?"
Essentially we need a "set" of files for each target, for the component tests it seems good enough. For the smaller tests you have spread out in the codebase I am not sure if they can be merged into 1 binary.  But I didn't spend time on it as the test was mostly a way to validate the build for now. I will take a closer look on the test portion.

I will add @evpobr changes and see if I can clean up the test section without touching the make files, also make sure it builds on Mac

CMakeLists.txt
```diff
@@ -0,0 +1,87 @@
...
+                           PUBLIC $<BUILD_INTERFACE:
+                                  ${CMAKE_CURRENT_SOURCE_DIR}/include
+                                  ${CMAKE_CURRENT_SOURCE_DIR}/celt
+                                  ${CMAKE_CURRENT_SOURCE_DIR}/silk
+                                  ${CMAKE_CURRENT_SOURCE_DIR}/silk/float
+                                  >
+                                  $<INSTALL_INTERFACE:include>)
+target_compile_definitions(opus PRIVATE USE_ALLOCA CUSTOM_MODES OPUS_BUILD)
+if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
+  target_link_libraries(opus m)
```
ras0219-msft: This should be `PRIVATE m`

CMakeLists.txt
```diff
@@ -0,0 +1,87 @@
...
+endif()
+
+# TODO: install lib, headers and tests
+
+# demo
+add_executable(opus_demo ${opus_demo})
+set_property(TARGET opus_demo PROPERTY C_STANDARD 99)
+target_link_libraries(opus_demo opus)
```
ras0219-msft: `PRIVATE opus`

CMakeLists.txt
```diff
@@ -0,0 +1,87 @@
...
+target_include_directories(opus
+                           PUBLIC $<BUILD_INTERFACE:
+                                  ${CMAKE_CURRENT_SOURCE_DIR}/include
+                                  ${CMAKE_CURRENT_SOURCE_DIR}/celt
+                                  ${CMAKE_CURRENT_SOURCE_DIR}/silk
+                                  ${CMAKE_CURRENT_SOURCE_DIR}/silk/float
+                                  >
+                                  $<INSTALL_INTERFACE:include>)
+target_compile_definitions(opus PRIVATE USE_ALLOCA CUSTOM_MODES OPUS_BUILD)
+if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
```
ras0219-msft: The ${} expansion around `CMAKE_SYSTEM_NAME` is not needed and could trigger double-expansion.
```cmake
if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
```
will do the right thing :)

CMakeLists.txt
```diff
@@ -0,0 +1,87 @@
...
+  target_link_libraries(opus m)
+endif()
+
+# TODO: install lib, headers and tests
+
+# demo
+add_executable(opus_demo ${opus_demo})
+set_property(TARGET opus_demo PROPERTY C_STANDARD 99)
+target_link_libraries(opus_demo opus)
+if(${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
```
ras0219-msft: ${} expansion not required.

CMakeLists.txt
```diff
@@ -0,0 +1,87 @@
...
+  target_compile_definitions(opus_demo PRIVATE _CRT_SECURE_NO_WARNINGS)
+endif()
+
+# tests
+add_executable(test_opus_api ${test_opus_api})
+set_property(TARGET test_opus_api PROPERTY C_STANDARD 99)
+target_link_libraries(test_opus_api opus)
```
ras0219-msft: `PRIVATE`

CMakeLists.txt
```diff
@@ -0,0 +1,87 @@
...
+if(${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
```
ras0219-msft: ${} expansion not required.

CMakeLists.txt
```diff
@@ -0,0 +1,87 @@
...
+  target_compile_definitions(test_opus_api PRIVATE _CRT_SECURE_NO_WARNINGS)
+endif()
+add_test(test_opus_api test_opus_api)
+
+add_executable(test_opus_encode ${test_opus_encode})
+set_property(TARGET test_opus_encode PROPERTY C_STANDARD 99)
+target_link_libraries(test_opus_encode opus)
```
ras0219-msft: `PRIVATE`

CMakeLists.txt
```diff
@@ -0,0 +1,87 @@
...
+target_compile_definitions(test_opus_encode PRIVATE _CRT_SECURE_NO_WARNINGS)
+if(${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
```
ras0219-msft: ${} expansion not required

CMakeLists.txt
```diff
@@ -0,0 +1,87 @@
...
+  target_compile_definitions(test_opus_encode PRIVATE _CRT_SECURE_NO_WARNINGS)
+endif()
+add_test(test_opus_encode test_opus_encode)
+
+add_executable(test_opus_decode ${test_opus_decode})
+set_property(TARGET test_opus_decode PROPERTY C_STANDARD 99)
+target_link_libraries(test_opus_decode opus)
```
ras0219-msft: PRIVATE

CMakeLists.txt
```diff
@@ -0,0 +1,87 @@
...
+target_compile_definitions(test_opus_decode PRIVATE _CRT_SECURE_NO_WARNINGS)
+if(${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
```
ras0219-msft: ${} expansion not required

CMakeLists.txt
```diff
@@ -0,0 +1,87 @@
...
+  target_compile_definitions(test_opus_decode PRIVATE _CRT_SECURE_NO_WARNINGS)
+endif()
+add_test(test_opus_decode test_opus_decode)
+
+add_executable(test_opus_padding ${test_opus_padding})
+set_property(TARGET test_opus_padding PROPERTY C_STANDARD 99)
+target_link_libraries(test_opus_padding opus)
+if(${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
```
ras0219-msft: ${} expansion not required

CMakeLists.txt
```diff
@@ -0,0 +1,87 @@
...
+target_link_libraries(test_opus_decode opus)
+target_compile_definitions(test_opus_decode PRIVATE _CRT_SECURE_NO_WARNINGS)
+if(${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
+  target_compile_definitions(test_opus_decode PRIVATE _CRT_SECURE_NO_WARNINGS)
+endif()
+add_test(test_opus_decode test_opus_decode)
+
+add_executable(test_opus_padding ${test_opus_padding})
+set_property(TARGET test_opus_padding PROPERTY C_STANDARD 99)
+target_link_libraries(test_opus_padding opus)
```
ras0219-msft: PRIVATE

CMakeLists.txt
xnorpx: right thanks!

xnorpx:
@jmvalin the unittests is not ideal for this situation. As they all have mains and need to built into separate binaries I cannot really think of a nice solution without refactor the structure of the test and merge them into one binary. Also create a test_sources.mk to keep the file lists there. But probably out of scope for this change.

berkus:
JFYI http://lists.qt-project.org/pipermail/development/2018-October/034023.html

@jmvalin what is that requirement that doesn't allow you to drop Makefile.am and all this ancient crud? If it's some VMS system, then I'm sure there's a cmake build for it too.

jmvalin:
> > I assume that by "break change tracking" you mean that things will break if someone changes the file list
> 
> @jvmalin as far as i know, cmake will take a conservative approach in case of using globs and will fully regenerate build system on each triggered build. While correctly generated cmake does this only when configuration changes.
> 
> That's an "acceptable" trade off here, as i understand.

Still trying to understand what you mean exactly here, so let me describe two scenarios:
1) Download Opus, build with cmake, then change one line in some source file, build again
2) Download Opus, build with cmake, add a new source file to one of the .mk files, build again

Will the "build again" step for 1) recompile everything, or just the file I changed? Will the build in 2) proceed properly? Will it recompile everything or just the file I added.

xnorpx:
@jmvalin 

1. It will do incremental build just fine.
2. CMake will not regenerate the solution automagically as it doesn't see any changes so the user that add/remove a file needs to do either:

- Delete solution and regenerate solution to pick up the new file or remove the file
- Make a change in CMakelists.txt to get CMake to pick the new changes

But this is just for the local development when using CMake, I see this as a small price to pay to get initial CMake support for Opus

@berkus VS2017 V.15.8.5 did not crash for me when git rm a file from the repo. The file only had a red mark in the solution and could of course not be opened.

CMakeLists.txt
```diff
@@ -13,7 +13,7 @@ message(STATUS "  CMAKE_HOST_SYSTEM_NAME: ${CMAKE_HOST_SYSTEM_NAME}")
 message(STATUS "  CMAKE_HOST_SYSTEM_VERSION: ${CMAKE_HOST_SYSTEM_VERSION}")
```
xnorpx: Make change to keep filepaths and folder structure

jmvalin:
> @jmvalin
> 
>     1. It will do incremental build just fine.
> 
>     2. CMake will not regenerate the solution automagically as it doesn't see any changes so the user that add/remove a file needs to do either:
> 
> 
>     * Delete solution and regenerate solution to pick up the new file or remove the file
> 
>     * Make a change in CMakelists.txt to get CMake to pick the new changes
> 
> 
> But this is just for the local development when using CMake, I see this as a small price to pay to get initial CMake support for Opus

OK, sounds reasonable (just wanted to make sure that 1 works fine).

berkus:
> VS2017 V.15.8.5 did not crash for me when git rm a file from the repo. The file only had a red mark in the solution and could of course not be opened.

@xnorpx what usually happens is this:
* cmake sees that there's a glob used 
* it then forces a regeneration of VS solution when you press Build
* build step re-generates the solution file from within VS
* VS goes crazy and crashes

If MS fixed this, then excellent and full-speed ahead.

jmvalin:
I'm having a hard time tracking down all the commits in this pull request (and honestly I don't actually use github much since this is just a mirror), so it would help to see this as a single commit. When it's ready, it would be good to post it on the list for review.

xnorpx:
@jmvalin yeah sorry for the spam, let me switch to another dev branch for my last changes, once ready I squash, clean up, force update this branch and ping this thread. If you want to avoid having open pull request you can close it and I reopen a new one once ready.

berkus:
Wait, how did this turn into Windows-only thing?? CMake is by definition cross-platform, adding it only for Windows makes no sense.

evpobr:
@jmvalin , not sure i understand the `MAY_HAVE` and `PRESUME` defines do, `MAY_HAVE` enables autodetection of SIMD sets? `PRESUME` builds for exact SIMD set?

xnorpx:
@berkus I was thinking first iteration Windows only as that I can test easiest. Then add Mac and Linux on top of that.

@evpobr Yeah the SIMD implementation is not clear to me and is were I am now.

jmvalin:
@evpobr MAY_HAVE means "the CPU may have that instruction set, but may not have it, so check at run-time", whereas PRESUME means "the CPU definitely has that instruction set, and no need to check at run-time"

jmvalin:
@xnorpx Opus should build as a single opus.lib library. No need to split into celt/silk (it used to be that way, but only for historical reasons)

evpobr:
> MAY_HAVE means "the CPU may have that instruction set, but may not have it, so check at run-time", whereas PRESUME means "the CPU definitely has that instruction set, and no need to check at run-time"

@jmvalin  Still don't understand. Why not enable all available instructions and check it runtime?

xnorpx:
@evpobr removed.

I do commits here: https://github.com/xnorpx/opus/tree/cmake-dev

Until it's ready for final review.

jmvalin:
> @jmvalin Still don't understand. Why not enable all available instructions and check it runtime?

Many people will build Opus for a specific application. They know exactly what architecture they target. For them, only including one architecture saves space (e.g. for embedded), not to mention that our runtime detection code is OS-specific, even if the optimizations themselves are not. So some people just can't have run-time detection, but they can have the optimizations hardcoded on. Again, this is especially true for embedded applications.

evpobr:
Ok. We can presume SSE, SSE2 for AMD64, because this CPU architecture always has these SIMD.

@xnorpx , i think both `MAY_BE` and `PRESUME` must be [dependent options](https://cmake.org/cmake/help/latest/module/CMakeDependentOption.html).

Something to start from (untested):

```
opus_detect_sse(HAVE_SSE)

cmake_dependent_option(OPUS_MAY_HAVE_SSE "Use SSE if available" ON "HAVE_SSE" OFF)
cmake_dependent_option(OPUS_PRESUME_SSE "Use SSE always (requires CPU with SSE)" ON "OPUS_MAY_HAVE_SSE;" OFF)
```

xnorpx:
@evpobr pushed dependent option changes for SSE

xnorpx:
@jmvalin do you have any examples or preference how install structure of lib/headers and tests should look like?

evpobr:
@xnorpx , see [here](https://packages.ubuntu.com/cosmic/amd64/libopus-dev/filelist) and other Opus packages. Do not install tests.

jmvalin:
@xnorpx I have a strong preference for "exactly where the other build system puts things". Otherwise we're going to run into problems with different people ending up with different setups because of the build system.

xnorpx:
@jmvalin squashed and made a forced push with 1 commit also updated the description for the pull request.

.gitignore
```diff
@@ -83,3 +83,5 @@ silk/tests/test_unit_LPC_inv_pred_gain
 src/Debug
 src/Release
 src/x64
+build
```
berkus: Use slashes as conventional in git when specifying directories in paths, e.g. `/build` or `build/`, otherwise it could match files.

.gitignore
```diff
@@ -83,3 +83,5 @@ silk/tests/test_unit_LPC_inv_pred_gain
...
+*.json
```
berkus: Why?

Remember that cmake uses out-of-src-dir-builds by default, so if you ignore a build folder, all generated json files are inside it and are already ignored. Blanket ignoring all json files may have unintended consequences.

.gitignore
evpobr: /*[Bb]uild*/

Ignores al commonl variants: `build`, `cmake-build`, `CMakeBuild` etc.

CMakeLists.txt
```diff
@@ -0,0 +1,128 @@
...
+  elseif(OPUS_PRESUME_SSE)
+    target_compile_definitions(objopus
+                               PRIVATE
+                               OPUS_X86_PRESUME_HAVE_SSE
+                               OPUS_X86_PRESUME_HAVE_SSE2
+                               OPUS_X86_PRESUME_SSE4_1)
+  endif()
+
+  if(NOT MSVC)
+    set(CMAKE_C_FLAGS "-msse4.1")
```
berkus: It makes sense to use a variant of the following macro, as not all compilers support these flags:

```
# Check compiler support for FLAG and set NAME_FLAG_SUPPORTED if supported.
# Add compiler flag to definitions if supported.
macro(check_and_set_flag NAME FLAG)
    include(CheckCXXCompilerFlag)
    check_cxx_compiler_flag(${FLAG} ${NAME}_FLAG_SUPPORTED)
    if (${NAME}_FLAG_SUPPORTED)
        add_definitions(${FLAG})
    endif()
endmacro()
```

Just replace add_definitions() with flag check and `set(CMAKE_C_FLAGS)` conditionally. Use [check_c_compiler_flag](https://cmake.org/cmake/help/latest/module/CheckCCompilerFlag.html?highlight=checkccompilerflag) for C sources, of course.

opus_config.cmake
```diff
@@ -0,0 +1,51 @@
...
+                       "Use SSE always (requires CPU with SSE)"
+                       ON
+                       "HAVE_SSE;OPUS_PRESUME_SSE"
+                       OFF)
+cmake_dependent_option(OPUS_MAY_HAVE_SSE
+                       "Use SSE if available"
+                       ON
+                       "HAVE_SSE;NOT OPUS_PRESUME_SSE"
+                       OFF)
+option(FIXED_POINT "Use fixed-point code (for device with less powerful FPU" NO)
```
berkus: Typo, `for devices`

berkus:
@xnorpx left a few comments, but otherwise it looks _excellent_. Well done, Marcus!

berkus:
Additionally, I have checked a build on **macOS High Sierra**, it compiled, but left this warning: some flag may be missing

```
~/R/o/_build_ ((d576b470…)|…) $ ninja
[14/150] Building C object CMakeFiles/objopus.dir/src/opus_decoder.c.o
../src/opus_decoder.c:37:10: warning: You appear to be compiling without optimization, if so opus will be very slow. [-W#pragma-messages]
# pragma message "You appear to be compiling without optimization, if so opus will be very slow."
         ^
1 warning generated.
```

evpobr:
> Additionally, I have checked a build on **macOS High Sierra**, it compiled, but left this warning: some flag may be missing
> 
> ```
> ~/R/o/_build_ ((d576b470…)|…) $ ninja
> [14/150] Building C object CMakeFiles/objopus.dir/src/opus_decoder.c.o
> ../src/opus_decoder.c:37:10: warning: You appear to be compiling without optimization, if so opus will be very slow. [-W#pragma-messages]
> # pragma message "You appear to be compiling without optimization, if so opus will be very slow."
>          ^
> 1 warning generated.
> ```

It was `Debug` build?

.gitignore
xnorpx: Agreed, removed

.gitignore
xnorpx: Updated based on suggestions

opus_config.cmake
xnorpx: fixed

berkus:
> > It was `Debug` build?
> 
> Most likely, when I build using:
> 
> "cmake .. -DCMAKE_BUILD_TYPE=Release -DBUILD_TESTING=1"
> 
> the pragma warning goes away as expected

Makes sense! Indeed, I just launched default `cmake -G Ninja ..`.

CMakeLists.txt
xnorpx: added check for SSE4 support

xnorpx:
@jmvalin here is a proposal for CI for the CMake build

Azure Devops Free Builds
https://azure.microsoft.com/en-us/blog/announcing-azure-pipelines-with-unlimited-ci-cd-minutes-for-open-source/
Config:
https://github.com/xnorpx/opus/blob/cmake-azuredevops/azure-buildpipeline.yaml
Builds:
https://xnorpx.visualstudio.com/opus%20build/_build/results?buildId=84&view=logs

Windows
 ARM
 X86
 X64
   Debug
   Release
Mac
 X64
   Debug
   Release
Linux
 X64
   Debug
   Release

NielsHygum:
In Android Studio the standard CMake is currently at version 3.6 and in Ubuntu 16 LTS the standard CMake version is 3.5... maybe the required cmake version could be changed to something less than 3.12?

xnorpx:
@NielsHygum the script is currently using features from 3.12 hence the required version. You can upgrade CMake for both Ubuntu and Android Studio.

https://stackoverflow.com/questions/51471908/upgrade-cmake-version-installed-with-android-studio-ubuntu-18-04

https://askubuntu.com/questions/355565/how-do-i-install-the-latest-version-of-cmake-from-the-command-line

While I understand this is annoying to update your dev/build environment I think it's a good trade off to use more modern version of CMake and Ubuntu and various tooling will eventually catch up.

xnorpx:
@evpobr list TRANSFORM

evpobr:
Iterate over list and make `string(STRIP ...)`? Available starting from 3.0.

xnorpx:
So first we probably should settle on a minimum requirement. Is 3.0 good or will it just be a matter of time before someone open a issue that they need 2.8 for their in-house buildsystem?

evpobr:
> So first we probably should settle on a minimum requirement. Is 3.0 good or will it just be a matter of time before someone open a issue that they need 2.8 for their in-house buildsystem?

CMake is about "It just works". So don't make people sad bumping CMake version without reason.

Believe me, i hate those guys who don't use important features like package config modules and stuck with `include_directories` etc just because they want to support some ancient Linux distro.

Some thoughts:

* 2.6 is dead. Don't use it.
* 2.8.11 lacks `target_include_directories`, so package config modules don't work properly. Do not use.
* 2.8.12 has everything you need. It was previous Ubuntu LTS default version. I'm pretty sure you can use it without `list TRANSFORM`.
* 3.0 is first of 3.x. That's fine to use it.
* 3.1 is good bugfix of 3.0. I use it most of the time.
* 3.2-3.3 Nothing really interesting
* Current Ubuntu LTS is 3.5.
* Android Studio users have 3.6.
* 3.10 will break Travis CI tests. It has 3.9 only.

So reasonable default is something between 3.0 and 3.5.

xnorpx:
Fair enough, I see what changes is needed to get under 3.5.

xnorpx:
Pushed downgrade to 3.1
Requirement tested locally with Ubuntu 18.04 with only CMake 3.1 installed
CI-Result: https://xnorpx.visualstudio.com/opus%20build/_build/results?buildId=85&view=logs

xnorpx:
Android also builds now: https://xnorpx.visualstudio.com/opus%20build/_build/results?buildId=96&view=logs

NielsHygum:
Thanks, great work! It compiles with out any problems in standard Android Studio on Linux Ubuntu.

xnorpx:
Added support for NEON on Android

xnorpx:
squashed changes into 1 commit.

xnorpx:
@jmvalin let me know if there is any concerns or anything missing to get this in now.

evpobr:
DLL_EXPORT must be private definition for Win32 shared build.

No VERSION & SOVERSION set, breaks ABI on Linux platform.

C_VISIBILTY_PRESET?

And I recommend do not create static/shared library targets. CMake is real cross-platform, on some platforms static or shared libraries are not available.

jmvalin:
@xnorpx One minor question: why are you adding an include guard in config.h? Currently it doesn't have any because it's only supposed to get included in C files (so it can't get included twice). Other than that, I'm OK with the non-cmake changes your patch makes. I can't really review the cmake files since I've almost never used it. So once you're happy with it, I would recommend sending the patch to the mailing list for wider review.

xnorpx:
@evpobr will check those changes and might ping you private for clarification

@jmvalin true will fix the config. Once @evpobr has signed off I will send patch to the mailing list.

evpobr:
> No VERSION & SOVERSION set, breaks ABI on Linux platform.

https://github.com/xiph/opus/blob/9791b22b2c83980f6b4386c870cad58557c78007/configure.ac#L25-L27

https://github.com/xiph/opus/blob/9791b22b2c83980f6b4386c870cad58557c78007/Makefile.am#L79


These lines are very important for any Unix-based system.

To use it with CMake you must:

1. Parse `configure.ac` and extract `OPUS_LT_CURRENT`, `OPUS_LT_REVISION` and `OPUS_LT_AGE` values
2. Set `OPUS_LIBRARY_VERSION_MAJOR` to `OPUS_LT_CURRENT` - `OPUS_LT_AGE`
3. Set `OPUS_LIBRARY_VERSION_MINOR` to `OPUS_LT_AGE` 
4. Set `OPUS_LIBRARY_VERSION_PATCH` to `OPUS_LT_CURRENT` 
5. Set `OPUS_LIBRARY_VERSION` to `${OPUS_LIBRARY_VERSION_MAJOR}.${OPUS_LIBRARY_VERSION_MINOR}.${OPUS_LIBRARY_VERSION_PATCH}`

You must get `OPUS_LIBRARY_VERSION` equal to `0.7.0`. CMake build will produce shared library with correct name `libopus.so.0.7.0`.

```cmake
set_target_properties(opus PROPERTIES
    SOVERSION ${OPUS_LT_VERSION}
    VERSION ${OPUS_LIBRARY_VERSION)}
```

More info:

https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html#Updating-version-info
http://blog.asleson.org/index.php/2014/07/08/libtool-library-versioning-version-info-currentrevisionage/
https://cmake.org/cmake/help/v3.1/prop_tgt/SOVERSION.html

xnorpx:
@evpobr perfect will read up and implement, thanks!

evpobr:
> DLL_EXPORT must be private definition for Win32 shared build.

> C_VISIBILTY_PRESET?

It is good practice to export only public functions from library. On Unix plafroms usually all functions are exported by default from static and shared library, on Win32 platform no functions exported by default from DLL.

Opus library uses compiler attributes approach. Any public function  is marked with `OPUS_EXPORT`.

On Win32 platform when `DLL_EXPORT` is defined it is expanded to `__declspec(dllexport)`. Note, this works for DLL only. So you need to check `BUILD_SHARED_LIBS` and `WIN32` variables are `ON` before setting `DLL_EXPORT` define.

For Unix it is expanded to `__attribute__ ((visibility ("default")))` for any public function. But you need to add `-fvisibility=hidden` argument to compiler. To do this set library property `C_VISIBILTY_PRESET` to `hidden`. To check compiler support you can look at [this](https://gitlab.kitware.com/cmake/cmake/blob/master/Modules/GenerateExportHeader.cmake).

xnorpx:
ping @evpobr

CMakeLists.txt
```diff
@@ -0,0 +1,184 @@
...
+  elseif(OPUS_PRESUME_NEON)
+    target_compile_definitions(opus
+                               PRIVATE
+                               OPUS_ARM_PRESUME_NEON
+                               OPUS_ARM_PRESUME_NEON_INTR)
+  endif()
+endif()
+
+include(GNUInstallDirs)
+install(TARGETS opus opus
```
evpobr: Duplicate opus targets

CMakeLists.txt
```diff
@@ -0,0 +1,184 @@
...
+                                 ${OPUS_LIBRARY_VERSION_MAJOR}
+                                 VERSION
+                                 ${OPUS_LIBRARY_VERSION})
+
+target_include_directories(opus
+                           PUBLIC $<BUILD_INTERFACE:
+                                  ${CMAKE_CURRENT_SOURCE_DIR}/include >)
+target_link_libraries(opus PRIVATE ${CMAKE_REQUIRED_LIBRARIES})
+
+target_include_directories(opus
```
evpobr: Duplicated command, see above.

CMakeLists.txt
```diff
@@ -0,0 +1,184 @@
...
+
+add_library(opus ${opus_sources} ${opus_sources_float})
+
+set_target_properties(opus
+                      PROPERTIES SOVERSION
+                                 ${OPUS_LIBRARY_VERSION_MAJOR}
+                                 VERSION
+                                 ${OPUS_LIBRARY_VERSION})
+
+target_include_directories(opus
```
evpobr: Missing `$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>`.

Don't forget to move `include(GNUInstallDirs)` before this command.

CMakeLists.txt
```diff
@@ -0,0 +1,184 @@
...
+                                  ${CMAKE_CURRENT_SOURCE_DIR}/include >)
+target_link_libraries(opus PRIVATE ${CMAKE_REQUIRED_LIBRARIES})
+
+target_include_directories(opus
+                           PUBLIC $<BUILD_INTERFACE:
+                                  ${CMAKE_CURRENT_SOURCE_DIR}/include >)
+target_link_libraries(opus PRIVATE ${CMAKE_REQUIRED_LIBRARIES})
+set_property(TARGET opus PROPERTY POSITION_INDEPENDENT_CODE ON)
+
+target_include_directories(opus
```
evpobr: `target_include_directories` with `PRIVATE` include dirs is not required here. You can set both `PRIVATE` and `PUBLIC` includes with one command.

xnorpx:
@evpobr thanks, some sloppiness from there. Pushed update.

CMakeLists.txt
```diff
@@ -0,0 +1,179 @@
...
+  PRIVATE $<BUILD_INTERFACE:
+          ${CMAKE_CURRENT_SOURCE_DIR}/include
+          ${CMAKE_CURRENT_SOURCE_DIR}/celt
+          ${CMAKE_CURRENT_SOURCE_DIR}/silk>)
+
+target_link_libraries(opus PRIVATE ${CMAKE_REQUIRED_LIBRARIES})
+
+target_compile_definitions(opus PRIVATE CUSTOM_MODES OPUS_BUILD)
+
+set_property(TARGET opus PROPERTY POSITION_INDEPENDENT_CODE ON)
```
evpobr: Remove it. It was required by object library. CMake will set it automatically for shared library.

CMakeLists.txt
```diff
@@ -0,0 +1,179 @@
...
+
+if(HAVE_NEON)
+
+  add_sources_group(opus celt ${celt_sources_arm_neon_intr})
+  add_sources_group(opus silk ${silk_sources_arm_neon_intr})
+
+  # silk arm neon depends on main_Fix.h
+  target_include_directories(
+    opus
+    PRIVATE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/silk/fixed>)
```
evpobr: No need for $<BUILD_INTERFACE:> in `PRIVATE` section. You may need it in `PUBLIC` or `INTERFACE` section to distinguish between build and install stages.

Also if you don't use generator expression, you can simply add directory as `silk/fixed`:

```cmake
target_include_directories(opus PRIVATE silk/fixed)
```

opus_config.cmake
```diff
@@ -0,0 +1,73 @@
...
+       OFF)
+option(USE_ALLOCA "Use alloca for stack arrays (on non-C99 compilers)" OFF)
+option(BUILD_PROGRAMS "Build programs" OFF)
+
+set_property(GLOBAL PROPERTY USE_FOLDERS ON)
+set_property(GLOBAL PROPERTY C_STANDARD 99)
+
+if(MSVC)
+  add_definitions(-D_CRT_SECURE_NO_WARNINGS)
+  string(REGEX
```
evpobr: Remove it. It is not requirement. CMake script must contain required stuff only.

CMakeLists.txt
```diff
@@ -0,0 +1,179 @@
...
+                               OPUS_HAVE_RTCD)
+  elseif(OPUS_PRESUME_NEON)
+    target_compile_definitions(opus
+                               PRIVATE
+                               OPUS_ARM_PRESUME_NEON
+                               OPUS_ARM_PRESUME_NEON_INTR)
+  endif()
+endif()
+
+install(TARGETS opus
```
evpobr: Not correct.

```cmake
RUNTIME DESTINATION	${CMAKE_INSTALL_BINDIR}
```

CMakeLists.txt
```diff
@@ -0,0 +1,179 @@
+cmake_minimum_required(VERSION 3.1)
+
+project(opus LANGUAGES C)
```
evpobr: You can remove `LANGUAGES`. `C` and `CXX` are enabled by default.

evpobr:
@xnorpx , missing parts

* `opus.pc` generation
* [Package configuration module](https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#package-configuration-file)

You can adapt my code from `libsndfile`.

And that's all. Good job!

PS Please don't squash commits yet. Hard to see what was changed.

CMakeLists.txt
```diff
@@ -0,0 +1,179 @@
...
+if(HAVE_SSE)
+
+  add_sources_group(opus celt ${celt_sources_sse})
+  add_sources_group(opus celt ${celt_sources_sse2})
+
+  if(SSE4_FLAG_SUPPORTED)
+    add_sources_group(opus celt ${celt_sources_sse4_1})
+    if(FIXED_POINT)
+      add_sources_group(opus silk ${silk_sources_fixed_sse4_1})
+    else(FIXED_POINT)
```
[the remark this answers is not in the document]
xnorpx: I see, changed everywhere

opus_config.cmake
xnorpx: just had some hope that it would encourage to fix the 280 L4 warnings :) so our fork would be easier to merge

CMakeLists.txt
xnorpx: If I use "LANGUAGES C" it will only check for c compiler and the generation will be faster, so will leave this one

CMakeLists.txt
```diff
@@ -1,28 +1,49 @@
...
+
+get_library_version(OPUS_LIBRARY_VERSION OPUS_LIBRARY_VERSION_MAJOR)
+message(STATUS "Opus library version: ${OPUS_LIBRARY_VERSION}")
+
+get_package_version(PACKAGE_VERSION)
+message(STATUS "Opus package version: ${PACKAGE_VERSION}")
+
+string(REGEX REPLACE "^v([0-9]+.[0-9]+\\.?([0-9]+)?).*" "\\1" PROJECT_VERSION ${PACKAGE_VERSION})
+
+project(Opus LANGUAGES C VERSION ${PROJECT_VERSION})
```
berkus: Hmm, I think project() command clears a lot of stuff, so it makes sense to put it first right after `cmake_minimum_required()` no?

evpobr: This is ugly,  but allows `project` command to set up `PROJECT_VERSION_MINOR` `PROJECT_VERSION_MINOR` and other versions variables.

Acording to documentation, `cmake_minimum_required` must be the first command to set policies properly. `project` command is not required to be next command.

berkus: Ok.

CMakeLists.txt
```diff
@@ -0,0 +1,229 @@
...
+endif()
+
+add_sources_group(opus silk ${silk_sources})
+add_sources_group(opus celt ${celt_sources})
+
+if(FIXED_POINT)
+  add_sources_group(opus silk ${silk_sources_fixed})
+  target_include_directories(
+    opus
+    PRIVATE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/silk/fixed>)
```
evpobr:
```suggestion
    PRIVATE silk/fixed)
```

evpobr: Not required, because install phase makes no sense for private headers.

CMakeLists.txt
```diff
@@ -0,0 +1,229 @@
...
+  target_compile_definitions(opus PRIVATE FIXED_POINT=1 DISABLE_FLOAT_API)
+else()
+  add_sources_group(opus silk ${silk_sources_float})
+  target_include_directories(
+    opus
+    PRIVATE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/silk/float>)
```
evpobr:
```suggestion
    PRIVATE silk/float)
```

xnorpx:
@evpobr is everything resolved now or I am missing something? I got time freed up again so we can try finish this.
@jmvalin I tried to mail the email list in December but still did not get my member ship approved and it does not look like my email with the patch went through either, how should I proceed with the email list?

opus_functions.cmake
```diff
@@ -0,0 +1,159 @@
...
+function(check_and_set_flag NAME FLAG)
+  include(CheckCCompilerFlag)
+  check_c_compiler_flag(${FLAG} ${NAME}_FLAG_SUPPORTED)
+  if(${NAME}_FLAG_SUPPORTED)
+    add_definitions(${FLAG})
+  endif()
+endfunction()
+
+include(CheckIncludeFile)
+function(opus_detect_sse HAVE_SSE)
```
evpobr: You detect SSE but check AVX flag? Also there are no SSE2, SSE4.1 and AVX headers\flags checks. BTW for x86_64 arch SSE and SSE2 must be presumed, they are present.

opus_functions.cmake
```diff
@@ -0,0 +1,159 @@
...
+    string(STRIP ${OPUS_PACKAGE_VERSION}, OPUS_PACKAGE_VERSION)
+    string(REPLACE \n
+                   ""
+                   OPUS_PACKAGE_VERSION
+                   ${OPUS_PACKAGE_VERSION})
+    string(REPLACE ,
+                   ""
+                   OPUS_PACKAGE_VERSION
+                   ${OPUS_PACKAGE_VERSION})
+    set(PACKAGE_VERSION ${OPUS_PACKAGE_VERSION} PARENT_SCOPE)
```
evpobr: You probably need convert functions in this file to macroses. Macros can set variables and they are in PARENT_SCOPE by default.

xnorpx: is there any point to convert everything to macros except to avoid the parent_scope flag?

evpobr:
No.

opus_functions.cmake
xnorpx: ok, I will leave it as functions (personal preference)

opus_functions.cmake
xnorpx: good points I will revisit this

xnorpx: I added detection for SSE1 and SSE2. 

I am not sure about defaulting any presume decision here, I feel more comfortable adding individual options for PRESUME SSE1,2,4 and the let the user decide this depending on his targets.

evpobr:
@xnorpx , please don't reformat. Damn hard to see important changes. It looks like all the files were changed.

xnorpx:
@evpobr yes sorry about that, itchy finger and didn't realize that I didn't commit my previous changes. Won't do it until final rebase now.

opus_functions.cmake
xnorpx: ok added

xnorpx:
ping @evpobr

CMakeLists.txt
```diff
@@ -0,0 +1,241 @@
...
+  PRIVATE ${CMAKE_CURRENT_BINARY_DIR}
+          ${CMAKE_CURRENT_SOURCE_DIR}
+          celt
+          silk)
+
+target_link_libraries(opus PRIVATE ${CMAKE_REQUIRED_LIBRARIES})
+
+target_compile_definitions(opus PRIVATE CUSTOM_MODES OPUS_BUILD)
+
+if(BUILD_SHARED_LIBS AND MSVC)
```
evpobr:
```suggestion
if(BUILD_SHARED_LIBS AND WIN32)
```

evpobr:
I've got linker errors. Can you test clean build?

````
[build]    Creating library opus.lib and object opus.exp
[build] encode_frame_FIX.c.obj : error LNK2001: unresolved external symbol _SILK_VAD_GETSA_Q8_IMPL
[build] find_LPC_FIX.c.obj : error LNK2001: unresolved external symbol _SILK_BURG_MODIFIED_IMPL
[build] burg_modified_FIX.c.obj : error LNK2001: unresolved external symbol _SILK_INNER_PROD16_ALIGNED_64_IMPL
[build] burg_modified_FIX_sse4_1.c.obj : error LNK2001: unresolved external symbol _SILK_INNER_PROD16_ALIGNED_64_IMPL
[build] opus.dll : fatal error LNK1120: 3 unresolved externals
[build] ninja: build stopped: subcommand failed.
[build] Build finished with exit code 1
````

**Comment by evpobr** [comment]
Many warnings in x86 mode:

````
[build] cl : Command line warning D9025 : overriding '/arch:SSE' with '/arch:SSE2'
[build] cl : Command line warning D9025 : overriding '/arch:SSE2' with '/arch:AVX'
````

**Comment by evpobr** [comment]
I suspect `MAY_HAVE` and `PRESUME` must be both set, you set only one of them.

File: opus_config.cmake [hunk 35]
```diff
@@ -0,0 +1,103 @@
...
+
+include(CMakeDependentOption)
+
+opus_detect_sse(COMPILER_SUPPORT_SIMD)
+if(COMPILER_SUPPORT_SIMD)
+  if(CMAKE_SIZEOF_VOID_P EQUAL 8)
+    set(OPUS_PRESUME_SSE1 ON)
+    set(OPUS_PRESUME_SSE2 ON)
+  else()
+    cmake_dependent_option(OPUS_PRESUME_SSE1
```
evpobr: `OPUS_PRESUME_SSE1` depends on itself?

xnorpx: Regarding linker error, can you send the cmake line you are using?

xnorpx: Good comments, please have another look.

evpobr:
> @evpobr I need to do some more testing, but regarding current state is it good to go?

I think it is good.

berkus:
What's up with merging it then?

jmvalin:
Considering that these patches don't really change existing files (unless I overlooked something), I think it's fine to all collapse them into a single patch. Can you do that and then post the result on the mailing list for review? I'd rather have that review on the list (more people paying attention -- including me) than on github.

xnorpx:
I managed to subscribe to the mailing list now, I guess spam filter killed the confirm email on work email. Once I verified builds I will rebase, give kudos to @evpobr in commit and create patch.

xnorpx:
The patch is on the mailing list so should be good for merge in a couple of days unless any other feedback comes in.

evpobr:
> The patch is on the mailing list so should be good for merge in a couple of days unless any other feedback comes in.

Great work @xnorpx. I we find bug later we will fix it.

xnorpx:
@jmvalin Thanks! Consider adding a CMake build on any platform on any buildsystem to sanity check the CMake integration.

berkus:
Yay, gratz! Welcome to 21st century, opus.

jmvalin:
Any thoughts about the VS2015 files in the opus/win32 directory? Are those eventually going to become useless in the longer term so they can be removed? I'm still leaving them for now, but it'd be nice if they could go away at some point.

xnorpx:
@jmvalin I would remove it asap, VS2019 was release this week. Change the Appveyor CI step to use CMake and then remove the win32 folder.

Nice to have would be to have static/dynamic builds for Windows/Mac/Linux and Android running on pull requests.

jmvalin:
Well, I'm not removing anything just yet -- at least not before people have had time to use the cmake setup and discover any issues it may have compared to the build files they've been using. The good news is that I'm planning on releasing 1.3.1 tomorrow or Friday, so we'll have the cmake stuff in the hands of users very soon. The only thing I'd like to fix at this point is defaulting to release builds (see email on the list).

PR #101, opened 2018-11-03
Make RNN faster by transposing weights

dofuuz:
Transpoing weights(changing memory layout) can speedup RNN about 2.5x.

↓ Before transposing
```
PS D:\work\opus\win32\VS2015\x64\Release> Measure-Command {.\opus_demo -e audio 16000 1 32000 rec2_2017_11_02_14_34_31.p
cm rec2_2017_11_02_14_34_31.opus}
libopus 1.3-dirty
Encoding 16000 Hz input at 32.000 kb/s in auto bandwidth with 320-sample frames.
average bitrate:              31.507 kb/s
maximum bitrate:              66.000 kb/s
active bitrate:               38.296 kb/s
bitrate standard deviation:    7.944 kb/s


Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 3
Milliseconds      : 517
Ticks             : 35172426
TotalDays         : 4.07088263888889E-05
TotalHours        : 0.000977011833333333
TotalMinutes      : 0.05862071
TotalSeconds      : 3.5172426
TotalMilliseconds : 3517.2426
```

↓ After transposing
```
PS D:\work\opus\win32\VS2015\x64\Release> Measure-Command {.\opus_demo -e audio 16000 1 32000 rec2_2017_11_02_14_34_31.p
cm rec2_2017_11_02_14_34_31.opus}
libopus 1.3-dirty
Encoding 16000 Hz input at 32.000 kb/s in auto bandwidth with 320-sample frames.
average bitrate:              31.507 kb/s
maximum bitrate:              66.000 kb/s
active bitrate:               38.296 kb/s
bitrate standard deviation:    7.944 kb/s


Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 3
Milliseconds      : 420
Ticks             : 34200928
TotalDays         : 3.95844074074074E-05
TotalHours        : 0.000950025777777778
TotalMinutes      : 0.0570015466666667
TotalSeconds      : 3.4200928
TotalMilliseconds : 3420.0928
```

It's about 2~3% speedup for opus encoding.

I also edited dumping scripts to transpose weights as needed. (Couldn't test it tho.)


jmvalin:
There's actually a good reason (at least in theory) why the weights are the way the are. The current ordering should actually be faster once the code is properly vectorized. The transposed version in this patch is probably easier to vectorize for the compiler (hence why it's faster), but once properly vectorized, I think the current ordering will be faster because we can avoid the horizontal sums at the end (which the transposed version would need). Feel like doing a bit of vectorization instead?

dofuuz:
Can you give some clue (keyword, link...) about optimization that you are going to use?
As I know, SIMD vectorization needs continuous memory access anyway.

jmvalin:
Basically, consider a matrix-vector product like this one:

   for (i=0;i<N;i++)
   {
      /* Compute update gate. */
      float sum = layer->bias[i];
      for (j=0;j<M;j++)
         sum += layer->input_weights[j*stride + i]*input[j];
      output[i] = WEIGHTS_SCALE*sum;
   }

If you're using SSE, you'll want to unroll the outer loop (on i) by 4, so that each vector computes four outputs and each load from the weight matrix loads elements from 4 lines at a time. If you don't unroll the inner loop (on j), then you load one input[] element at a time and broadcast it to all vector elements before the multiply. However, if you want to go a little faster, then you can unroll the inner loop by 4 as well and just load once (4 values) and broadcast 4 times. You can also keep 2 or 4 separate partial sums so you don't get hit by the add latency. Get the general idea?

dofuuz:
Created new PR #103

PR #104, opened 2018-11-12
replace a few WIN32 checks with _WIN32

sezero:
also remove a left-over windows-or-linux check from
within a windows-only case in debug.c.

### DIFF
--- a/celt/stack_alloc.h
+++ b/celt/stack_alloc.h
@@ -40,7 +40,7 @@
 #endif
 
 #ifdef USE_ALLOCA
-# ifdef WIN32
+# ifdef _WIN32
 #  include <malloc.h>
 # else
 #  ifdef HAVE_ALLOCA_H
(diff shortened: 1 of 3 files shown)


sezero:
PING ?

jmvalin:
What's the difference between WIN32 and _WIN32 exactly? (not a Windows user) Also, I'm a bit concerned about the changes to silk/debug.c which is not used by default (so you wouldn't notice if it gets broken).

sezero:
> What's the difference between WIN32 and _WIN32 exactly? (not a Windows user)

`_WIN32` is a compiler define, both for gcc and visual studio. WIN32 w/o the _ prefix
is just a user define. 

> Also, I'm a bit concerned about the changes to silk/debug.c which is not used by
> default (so you wouldn't notice if it gets broken).

And the current situation is confusing (to me, at least..)

jmvalin:
> And the current situation is confusing (to me, at least..)

What's confusing to you?

sezero:
> > And the current situation is confusing (to me, at least..)
> 
> What's confusing to you?

Trying to understand the hidden meaning in duplicated ifdefs where there is none.

PR #107, opened 2018-12-21
API for checking whether the encoder is in DTX

gustafullberg:
In WebRTC, we would like to be able to differentiate between "normal" frames and the comfort noise update frames that are encoded before and in-between DTX frames. These CNG update frames could then be flagged for routing or statistics reasons. Such flagging could also make it easier for a jitterbuffer to differ DTX from packet loss.

This PR implements an encoder CTL named OPUS_GET_IN_DTX. OPUS_GET_IN_DTX returns 1 if the last encoded frame was either a DTX frame, or a CNG update frame that is sent every 420 ms between the DTX frames.

This is achieved by checking if the number of consecutive inactive frames (nb_no_activity_frames) is greater or equal to NB_SPEECH_FRAMES_BEFORE_DTX.

### DIFF
--- a/include/opus_defines.h
+++ b/include/opus_defines.h
@@ -168,6 +168,7 @@ extern "C" {
 /* Don't use 4045, it's already taken by OPUS_GET_GAIN_REQUEST */
 #define OPUS_SET_PHASE_INVERSION_DISABLED_REQUEST 4046
 #define OPUS_GET_PHASE_INVERSION_DISABLED_REQUEST 4047
+#define OPUS_GET_IN_DTX_REQUEST              4049
 
 /** Defines for the presence of extended APIs. */
 #define OPUS_HAVE_OPUS_PROJECTION_H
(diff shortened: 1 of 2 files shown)


mark4o:
DTX can be activated not only based on nb_no_activity_frames, but also within the SILK encoder based on SILK's voice activity detection.  See the inDTX member of silk_encoder_state.

gustafullberg:
Added a commit to the PR to address the comment by @mark4o.
The OPUS_GET_IN_DTX should now properly handle both Opus DTX and Silk DTX.
Please take a look @mark4o and @hlundin.

gustafullberg:
Ping @jmvalin @mark4o 

We (WebRTC and Google) would love to see this main-lined.
Is this something we could get into master? Any further concerns?

jmvalin:
Sorry for the delay. Can you post the latest version of the patch to the mailing list so that everyone can have a look at it?

gustafullberg:
Posted to the list:
http://lists.xiph.org/pipermail/opus/2019-April/004299.html

gustafullberg:
Fixed detection of Silk DTX for stereo.

gustafullberg:
I squashed the four commits to make the PR suitable for merging.

jmvalin:
Are you talking about the patch I already merged or is that a new one?

gustafullberg:
Sorry, I didn't realize you merged the patch :)
Thanks. Closing the pull request.

PR #109, opened 2019-01-05
mips:  Rename MIPSr1_ASM to MIPSr2dsp_ASM to reflect true requirements.

speachy:
All of the mips code in opus requires a mips32r2 core with dsp or dsp2 extensions, so
update the #define to document this appropriately.

I did not rename the various header files that have "mipsr1" in their name, as that seemed like unnecessary churn. 

Signed-off-by: Solomon Peachy <pizza@shaftnet.org>

### DIFF
--- a/Makefile.mips
+++ b/Makefile.mips
@@ -46,7 +46,7 @@ ldlibs-from-libs        = $(addprefix -l,$(1))
 
 WARNINGS = -Wall -W -Wstrict-prototypes -Wextra -Wcast-align -Wnested-externs -Wshadow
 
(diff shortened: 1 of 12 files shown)


jmvalin:
Can you contact the original authors of the patch to see what they have to say about it. I don't really have any way to test that arch.

neheb:
Does it also require hardware floating point?

AleXoundOS:
What's the correct way of building opus for mips32r2 (mipsel, little-endian)? How to test this?

PR #117, opened 2019-04-10
CMake changes - Make release build default, made CUSTOM_MODE an optio…

xnorpx:
CMake changes - Make release build default, made CUSTOM_MODE an option with default off and added missing buildflags for Linux and security.


xnorpx:
Ping @evpobr @jmvalin

Issue #132, opened 2019-05-01
CMake - Shouldn't tamper march, /arch settings


Chocobo1:
AFAIK opus has some built-in detection for available CPU instructions and will select/use them automatically, so explicitly setting `march` (gcc) and `/arch` (msvc) is not required at all.
Furthermore, setting them is harmful for distro package maintainers: for example opus compiled on a modern machine supporting AVX and later some user downloads the package on an older machine without AVX, running the program will crash badly.

https://github.com/xiph/opus/blob/master/opus_functions.cmake#L105-L178
https://hydrogenaud.io/index.php/topic,117526.msg970503.html#msg970503

xnorpx:
Looks like it: https://randomascii.wordpress.com/2016/12/05/vc-archavx-option-unsafe-at-any-speed/

Chocobo1:
>Looks like it: https://randomascii.wordpress.com/2016/12/05/vc-archavx-option-unsafe-at-any-speed/

http://www.virtualdub.org/blog/pivot/entry.php?id=347

Here is another blog post from the net, the second part is basically solving the same problem as the one you linked. The first part (crashing on old CPUs) is what I hope to raise in this issue, IIRC even a simple hello world program could crash when the arch setting is not right.

simoncent:
I'm encountering this exact problem right now with the Google Android emulator. There are options provided around this by the CMakeLists.txt file too, which don't seem to do anything! e.g. setting `-DOPUS_X86_MAY_HAVE_AVX=OFF` on the command line doesn't do anything because the flag is hardcoded here `check_and_set_flag(AVX -mavx)` for every file when `immintrin.h` is found. 

The only way I've managed to disable AVX is by deleting that line from `opus_functions.cmake`.

xnorpx:
I have sent patches to the Opus mail list, but no reply so far. Will ping again

hanseuljun:
Thanks a lot for all your work guys. I just wanted to provide a real example to support this direction. HoloLens (a $3000 AR headset) does not have support AVX...

xnorpx:
this has been merged to master

xnorpx:
@Chocobo1 @rillian this issue can be closed

Chocobo1:
Thanks!

PR #135, opened 2019-05-09
CMake: Disable AVX flag on MSVC to support older Processors

xnorpx:
Disable AVX flag #132


Chocobo1:
What about removing `check_and_set_flag(AVX -mavx)` at line 162? although it is not msvc reign the same logic of this PR applies.

Chocobo1:
Also I see there is this compile-time define: `OPUS_HAVE_RTCD`, configure.ac state it stands for "run-time CPU capabilities detection", if that's correct, then there shouldn't be any need for setting `march`, `/arch` (or even other `OPUS_X86_PRESUME_*`) at compile-time, the detection can all be done at runtime.

PR #136, opened 2019-05-10
CMake: Add support for detecting the presence of lrint(3) and lrintf(3)

teknoman117:
Detect the presence of lrint(3) and lrintf(3) in the CMake build system. Resolves warnings when building for ARM as they obviously don't support SSE, so it falls back on lrint and lrintf.


teknoman117:
Is anyone able to take a look at this?

teknoman117:
closing due to merging into xnorpx fork (I'm guessing they are the CMake build system maintainer).

PR #141, opened 2019-08-22
Close warning in opus_encoder.c while enable fuzzing

willson-chen:
stereo_voice_threshold, stereo_music_threhold and mode_threshold, these var only used in fuzzing.

### DIFF
--- a/celt/rate.c
+++ b/celt/rate.c
@@ -348,6 +348,7 @@ static OPUS_INLINE int interp_bits2pulses(const CELTMode *m, int start, int end,
             /*This if() block is the only part of the allocation function that
                is not a mandatory part of the bitstream: any bands we choose to
                skip here must be explicitly signaled.*/
+#ifndef FUZZING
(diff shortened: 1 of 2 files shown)


willson-chen:
There are two warnings left.
```
celt/rate.c:251:82: warning: unused parameter ‘prev’ [-Wunused-parameter]
       int *ebits, int *fine_priority, int C, int LM, ec_ctx *ec, int encode, int prev, int signalBandwidth)
                                                                                  ^~~~
celt/rate.c:251:92: warning: unused parameter ‘signalBandwidth’ [-Wunused-parameter]
       int *ebits, int *fine_priority, int C, int LM, ec_ctx *ec, int encode, int prev, int signalBandwidth)
                                                                                            ^~~~~~~~~~~~~~~
```
These two input var can't be excluded unless modifing the function declaration.

jmvalin:
Seems to be fixed. Please re-open if not

PR #142, opened 2019-08-29
enable test under cmake

willson-chen:
Enable test by add `include(CTest)` in CMakeLists.txt


xnorpx:
Enabling CTest makes the project messy so it was avoided, we can add another Option variable to make it more verbose instead of BUILD_TESTING

willson-chen:
IMHO, add `-DBUILD_TESTING=1` when cmake build is a more common way. So close this PR.

PR #144, opened 2019-09-05
Enable building opus_demo and opus_compare under cmake

willson-chen:
Enable building opus_demo and opus_compare under cmake by set BUILD_OPUS_PROGRAM on.


xnorpx:
It is set to OFF as we by default only build the library, you enable it with -D option in your buildsystem

willson-chen:
> It is set to OFF as we by default only build the library, you enable it with -D option in your buildsystem

agreed

PR #145, opened 2019-09-06
fixed doc/build_draft.sh run error

willson-chen:


rillian:
Thanks for the fix!

I went through and addressed several other issues on the [draft](https://github.com/xiph/opus/commits/draft) branch.

There's a still a problem with including `opus_source.base64`. Current xml2rfc calls `lxml.etree.parse` on the included file which wants it to have a root node. That can be solved by moving the enclosing `<t>` tag into `opus_source.base64`. However, it seems `<spanx style="raw">` is no longer supported, so the text ends up decorated with underscores or marked italic, etc., which breaks extracting the tarball again.

If you'd like to work on this some more, it would be great to find a fix that. Maybe just by restoring support in [xml2rfc](https://pypi.org/project/xml2rfc/)?

willson-chen:
> Thanks for the fix!

My pleasure. And shall I close this PR? Or wait for it being merged into master branch?

> If you'd like to work on this some more, it would be great to find a fix that. Maybe just by restoring support in xml2rfc?

OK. I will try. :)

rillian:
Let's leave it open until it's merged.

rillian:
Seems we've not made much progress here. I've merged this patch into master as 3cb71dbefd5a0a2ec4d692c6dd71c9ed6d207ffb and updated the `draft` branch. I'll close this PR; please open a new one if you make progress on the outstanding issues.

PR #147, opened 2019-09-30
CMake: add option to set BUILD_SHARED_LIBS variable

davidebeatrici:
From https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html:

> Global flag to cause [add_library()](https://cmake.org/cmake/help/latest/command/add_library.html#command:add_library) to create shared libraries if on.
>
> If present and true, this will cause all libraries to be built shared unless the library was explicitly added as a static library. This variable is often added to projects as an [option()](https://cmake.org/cmake/help/latest/command/option.html#command:option) so that each user of a project can decide if they want to build the project using shared or static libraries.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,6 +18,7 @@ message(STATUS "Opus project version: ${PROJECT_VERSION}")
 project(Opus LANGUAGES C VERSION ${PROJECT_VERSION})
 include(opus_buildtype.cmake)
 
+option(OPUS_BUILD_SHARED_LIBRARY "Build shared library" OFF)
 option(OPUS_STACK_PROTECTOR "Use stack protection" ON)
 option(OPUS_USE_ALLOCA "Use alloca for stack arrays (on non-C99 compilers)" OFF)
 option(OPUS_CUSTOM_MODES "Enable non-Opus modes, e.g. 44.1 kHz & 2^n frames"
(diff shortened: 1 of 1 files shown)


neheb:
What's wrong with just passing -DBUILD_SHARED_LIBS=ON ?

davidebeatrici:
I'm using this repository as submodule and building Opus in my own CMake project.

Enabling `BUILD_SHARED_LIBS` would cause all libraries in the project to be built as shared (except the ones explicitly added as a static).

This is what I'm doing right now (with this pull request applied):
```cmake
option(OPUS_BUILD_SHARED_LIBRARY "" ON)
add_subdirectory("${CMAKE_CURRENT_SOURCE_DIR}/opus-src")
target_link_libraries(my_target PRIVATE opus)
```

ZeroAbility:
This would be helpful for Windows builds. Please add, and thank you.

xnorpx:
@ZeroAbility this is in master already. @davidebeatrici please close this pull request (patches goes throw offical repo at xiph.org)

xnorpx:
https://github.com/xiph/opus/commit/c487f53ca59507d6d511aca4a3edc1d768633d19#diff-af3b638bc2a3e6c650974192a53c7291

davidebeatrici:
Sorry, I missed that. Thank you very much for pointing it out!

PR #149, opened 2019-10-06
configure: adjust x86 get cpu info inline assembly method for PIC case

sezero:
With old Darwin gcc versions 4.0.1 and 4.2.1, I get:
`checking How to get X86 CPU Info... configure: error: no supported Get CPU Info method, please disable run-time CPU capabilities detection or intrinsics`

config.log has:
```
conftest.c:41: error: can't find a register in class 'BREG' while reloading 'asm'
conftest.c:41: error: 'asm' operand has impossible constraints
```
Adjusting the tested code to deal with PIC case (just like the way it
is done in celt/x86/x86cpu.c) cures it.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -691,6 +691,18 @@ AS_IF([test x"$enable_intrinsics" = x"yes"],[
                  unsigned int CPUInfo2;
                  unsigned int CPUInfo3;
                  unsigned int InfoType;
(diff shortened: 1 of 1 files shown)


sezero:
PING?

sezero:
PING?

sezero:
PING ?

jmvalin:
Merged

PR #150, opened 2019-10-11
Silk decoder and DTX: avoid abrupt changes due to comfort noise refresh packets

jvicentep:
In this pull request the behavior of the Silk Decoder during DTX is improved:
 - First, the commit 4f5557c3095a1d212161609ff638cdae67a9b303: “Attenuate SILK PLC gain only for unvoiced speech” is reverted. With the use of the OPUS VAD for deciding when to go to DTX, there are occasions when frames with higher energy than the comfort noise are labeled as TYPE_NO_VOICE_ACTIVITY. A typical example is in the refresh packets that are sent during DTX regions. The current decoder was keeping the energy of the concealed signal which was making the noise during DTX too high. 
- Avoid abrupt changes due to comfort noise refresh packets. We have realized that the refresh packets are creating abrupt noise pumps when they represent a frame that is very different than the comfort noise. This is a common situation when the Silk VAD does not agree with the Opus VAD. In this pull request we propose to use those refresh packets for updating the comfort noise model and keep producing comfort noise during those 20 ms. That makes a smoother transition to the new comfort noise level.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -186,7 +186,7 @@ target_link_libraries(opus PRIVATE ${OPUS_REQUIRED_LIBRARIES})
 target_compile_definitions(opus PRIVATE OPUS_BUILD ENABLE_HARDENING)
 
 if(NOT MSVC)
-  target_compile_definitions(opus PRIVATE FORTIFY_SOURCE=2)
+  target_compile_definitions(opus PRIVATE _FORTIFY_SOURCE=2)
 endif()
 
 # It is strongly recommended to uncomment one of these VAR_ARRAYS: Use C99


jvicentep:
In the next figures we can see a couple of examples of the effect of this pull request: 
opus_dtx_improvement

another_comparison

boborjan2:
Reverting [4f5557c](https://github.com/xiph/opus/commit/4f5557c3095a1d212161609ff638cdae67a9b303) results in very high level of noise during DTX when the signal encoded is a continuous 1kHz beep that is manually muted/unmuted to trigger dtx.

PR #153, opened 2019-10-30
set stop when last frame

lulu920819:
Change code in opus_demo.c
It won't set stop=1 when last frame has the same size of the frame_size, which means `curr_read+remaining = frame_size` and `remaining = 0`. And in this situation, it encode an extra all 0 frame

### DIFF
--- a/src/opus_demo.c
+++ b/src/opus_demo.c
@@ -689,7 +689,7 @@ int main(int argc, char *argv[])
                 s=((s&0xFFFF)^0x8000)-0x8000;
                 in[i+remaining*channels]=s;
             }
-            if (curr_read+remaining < frame_size)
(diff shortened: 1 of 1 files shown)


mark4o:
The code only reads up to frame_size samples per channel into the buffer.  If less than the requested amount could be read, stop is set to 1.  The proposed change sets stop to 1 even when it was able to read the full requested amount.  That means that this condition will always be true and it will stop after reading the first frame.  Since the code is intended to read all of the frames, this change is not correct.

Because Opus has lookahead, a logical position would be that opus_demo -e should generate enough output frames to cover the input plus the lookahead, to avoid cutting off the final few milliseconds of audio.  That would mean that even if it detected that a full frame that it just read was the last frame of PCM input, it it should not set stop to 1 at that time.

PR #156, opened 2019-12-04
Bugfix: OPUS_GET_IN_DTX and Silk mid/side coding

gustafullberg:
The current implementation of OPUS_GET_IN_DTX does not properly handle the case when only the middle channel is encoded in Silk stereo.

This PR fixes the bug by only looking at noSpeechCouter for the second channel (side channel) if the previous frame was not "only middle".

The bug was found by a failing WebRTC test case:
https://webrtc-review.googlesource.com/c/src/+/159709


xnorpx:
@rillian looks like this was merged in 93b373e884915e802188cec844a7dda2fa184ebc and can be closed @gustafullberg (cleanup) ;)

rillian:
Indeed! Thanks.

PR #157, opened 2019-12-05
cmake: use PACKAGE_VERSION for the pkg-config file

DeadSix27:
The old variable `OPUS_LIBRARY_VERSION`, does not line up with what's used in [autoconf](https://github.com/xiph/opus/blob/master/configure.ac#L6), which uses the `PACKAGE_VERSION` instead, which in turn lines up with what version the projects `opusfile` and `libsndfile` check for, for example.
While at it, I also cleaned up the accidental double up in code.

Here are code examples:
[opusfile](https://github.com/xiph/opusfile/blob/master/configure.ac#L104) checks for >= 1.0.1
[libsndfile](https://github.com/erikd/libsndfile/blob/master/configure.ac#L354) checks for >= 1.1
Both checks would fail on `LIBRARY_VERSION` which is just `0.8.0`, but would work fine on `PACKAGE_VERSION` which is `1.3.1-4-gad8fe90d`


davidebeatrici:
Landed as 6d29f51a40be64e03acc2619f35b4971ba00617c.

xnorpx:
@DeadSix27  this has been merged

@rillian @mark4o  so this PR can be closed

rillian:
Landed in 6d29f51a4. Thanks!

PR #166, opened 2020-02-07
Fixes to the the activity flag that is passed to the Silk encoder

jvicentep:
The activity flag that is used for deciding when to go to DTX is passed to the Silk encoder. However, after that flag is passed, it is further modified. This pull request correct that behavior by introducing those modifications before calling the Silk encoder.

### DIFF
--- a/src/opus_encoder.c
+++ b/src/opus_encoder.c
@@ -892,34 +892,15 @@ static opus_val32 compute_frame_energy(const opus_val16 *pcm, int frame_size, in
 #endif
 
 /* Decides if DTX should be turned on (=1) or off (=0) */
-static int decide_dtx_mode(float activity_probability,    /* probability that current frame contains speech/music */
-                           int *nb_no_activity_frames,    /* number of consecutive frames with no activity */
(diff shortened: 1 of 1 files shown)


src/opus_encoder.c
```diff
@@ -1668,7 +1675,6 @@ opus_int32 opus_encode_native(OpusEncoder *st, const opus_val16 *pcm, int frame_
     if (st->mode != MODE_CELT_ONLY)
     {
         opus_int32 total_bitRate, celt_rate;
-        opus_int activity;
```
felicialim: could drop `activity = VAD_NO_DECISION` below now

jvicentep: Done. Thanks for pointing that out.

felicialim:
Thanks for catching and fixing this. Merged.

xnorpx:
if this is merged can this pr be closed @jvicentep ?

xnorpx:
@rillian this one seems to be merged as well

rillian:
Agreed.

PR #168, opened 2020-02-19
Fix celt decoder assertion when using OPUS_CUSTOM

zanshi:
When using OPUS_CUSTOM, `CELTDecoder->end` can be larger than 21.
This causes builds with `CUSTOM_MODES` and either `ENABLE_HARDENING` or `ENABLE_ASSERTIONS` to crash. 

Assert against 25 (maximum number of Bark bands) instead in OPUS_CUSTOM builds.


hoene:
Does this relate to https://github.com/xiph/opus/issues/172 ?

jmvalin:
Thanks for the patch. It's merged now.

PR #174, opened 2020-04-24
doc/build_draft.sh: fix run failed under dash

willson-chen:
[[ ]], the compound command is not supported by all shell interpreter. [ ], the buildin command is more common.


willson-chen:
Another way to fix this issue is to set the shell interpreter as bash

```bash
#!/bin/bash -e
```

petterreinholdtsen:
Sure, but sticking to bourne shell syntax seem like a better idea, for portability.  Not all unix like operating systems have /bin/bash.  Because of this, I recommend replacing [[ ]] with [ ] as the pull request proposes.

rillian:
pushed as 86e9bfc8bb759004bf6eaed19dabe06ee9c44e7b. Thanks!

PR #178, opened 2020-05-04
doc/draft-ietf-codec-opus.xml: fix typo

willson-chen:
Fix from 'exlcude' to 'exclude' in some toc value.
In fact, this typo will also result in xml2rfc failure.


willson-chen:
Some modification has merge into `draft` branch. Thanks for the reminder from @rillian. :)

PR #179, opened 2020-05-05
doc/draft-ietf-codec-opus.xml: fix anchor error

willson-chen:
There should not be space in the value of anchor.
Anchor values and correspoding xref targets are updated.


rillian:
Thanks for taking a look at this, but you seem to be duplicating the work I already did on the [draft branch](https://github.com/xiph/opus/commits/draft). I suggest building on top of that so you can just work on fixing new issues. Then add to your original pull request.

willson-chen:
Thanks for your reminder. :)

I didn't notice this branch, and made a big mistake.

PR #182, opened 2020-05-29
cmake/CFeatureCheck.cmake: fix feature tests failing when Opus is a submodule

davidebeatrici:
`CMAKE_SOURCE_DIR` corresponds to the top project's source directory.
`CMAKE_BINARY_DIR` corresponds to the top project's binary directory.

The usage of these variables doesn't cause any problems when Opus is built as a standalone project.

This is not the case when Opus is added as submodule: the variables are set by the project that calls `add_subdirectory()`.

The fix consists in using `PROJECT_SOURCE_DIR` and `PROJECT_BINARY_DIR`, which always refer to the current project.

### DIFF
--- a/cmake/CFeatureCheck.cmake
+++ b/cmake/CFeatureCheck.cmake
@@ -27,7 +27,7 @@ function(c_feature_check FILE)
 
   if (NOT DEFINED COMPILE_${FEATURE})
       message(STATUS "Performing Test ${FEATURE}")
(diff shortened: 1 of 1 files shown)


xnorpx:
@davidebeatrici  can you make the pull request in https://github.com/xnorpx/opus and I add it in the next set of cmake patches there might be some more CMAKE_SOURCE_DIR in that branch as well

davidebeatrici:
Done.

mark4o:
Thanks.  This has been merged in ce9d83be806941d48dc37034c0c6522a4bf09bb6.

davidebeatrici:
Excellent, thank you!

PR #183, opened 2020-06-11
CFeatureCheck: Use CMAKE_CURRENT_SOURCE_DIR, not CMAKE_SOURCE_DIR

nevyn:
The latter gets the directory of the outermost CMakeLists.txt. In the case where
libopus is submoduled into another project and libopus is add_subdirectory'd, this
means feature check will look in the parent project instead of this project's 'cmake'
folder for a feature check c file. Changing to the former fixes this.

### DIFF
--- a/cmake/CFeatureCheck.cmake
+++ b/cmake/CFeatureCheck.cmake
@@ -27,7 +27,7 @@ function(c_feature_check FILE)
 
   if (NOT DEFINED COMPILE_${FEATURE})
       message(STATUS "Performing Test ${FEATURE}")
-      try_compile(COMPILE_${FEATURE} ${CMAKE_BINARY_DIR} ${CMAKE_SOURCE_DIR}/cmake/${FILE}.c)
+      try_compile(COMPILE_${FEATURE} ${CMAKE_BINARY_DIR} ${CMAKE_CURRENT_SOURCE_DIR}/cmake/${FILE}.c)
   endif()
 
   if(COMPILE_${FEATURE})


xnorpx:
https://github.com/xiph/opus/pull/182

is getting merged to master soon. That should solve this as well?

davidebeatrici:
Fixed in #182 (now merged).

I decided to use `PROJECT_SOURCE_DIR` rather than `CMAKE_CURRENT_SOURCE_DIR` because the latter expands to the current [CMake subdirectory](https://cmake.org/cmake/help/latest/command/add_subdirectory.html)'s source directory.

In this case `CMAKE_CURRENT_SOURCE_DIR` works, but it wouldn't if `c_feature_check()` was called in a subdirectory.

mark4o:
@nevyn https://github.com/xiph/opus/pull/182 was merged in ce9d83be806941d48dc37034c0c6522a4bf09bb6; does that resolve the issue for you?

xnorpx:
There is a couple of bugs regarding correct opus version parsing when used as a module. I am working on change + test to make sure that this works well. Will try to get this ready for review during the weekend.

nevyn:
> 
> 
> @nevyn #182 was merged in [ce9d83b](https://github.com/xiph/opus/commit/ce9d83be806941d48dc37034c0c6522a4bf09bb6); does that resolve the issue for you?

Oh damn, I should've checked the outstanding PRs before making my own! Sorry for the redundancy. #182 looks like  an even better fix than mine. Thanks!!

davidebeatrici:
No problem!

Issue #186, opened 2020-07-07
Can't cross-compile on Mac to create universal binaries


andrewgrant:
It isn't currently possible with Opus to cross compile on Mac in order to create universal libraries that are capable of supporting both Intel and Apple silicon Mac's

danra:
I found this out the hard way, by realizing the libopus fat binary produced by CMake using the XCode generator was missing vector optimizations not just for the arm64 slice (which indeed doesn't support SSE, AVX) but also for the x86_64 slice :(

Perhaps prior to CMake 3.19 building UBs just completely didn't work; With CMake 3.19 it does work, but it doesn't produce the expected optimizations.

It may be beneficial to add a warning when trying to do universal builds until this is fixed.

xnorpx:
I don't have a M1 Mac so can't test it.

But pull requests is appreciated.

bear101:
With Xcode 13 you can build "multi-arch" binaries (universal binaries), i.e. you can compile with "-arch arm64 -arch x86_64".

Here's how I use CMake ExternalProject to build an OPUS universal binary: https://github.com/BearWare/TeamTalk5/blob/apple_m1/Library/TeamTalkLib/build/opus/CMakeLists.txt#L42

Here's a long thread were we discuss OPUS on Apple M1 using Rosetta emulator:
https://github.com/BearWare/TeamTalk5/issues/1067

PR #187, opened 2020-07-07
Add support for cross-compiling on Mac #186

andrewgrant:
Adds support for cross-compiling to different targets when building on Mac.

E.g the following builds Opus for an arm64 Mac target (Apple silicon)

./configure --host=aarch64-apple-macos make && make install

This allows building Opus for both architectures and using lipo to create a universal binary.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -9,12 +9,27 @@ lib_LTLIBRARIES = libopus.la
 
 DIST_SUBDIRS = doc
 
+AM_LDFLAGS =
+AM_CFLAGS =
 AM_CPPFLAGS = -I$(top_srcdir)/include -I$(top_srcdir)/celt -I$(top_srcdir)/silk \
               -I$(top_srcdir)/silk/float -I$(top_srcdir)/silk/fixed $(NE10_CFLAGS)
 
 include celt_sources.mk
 include silk_sources.mk
 include opus_sources.mk
+
+# Pass x86_64 in compiler and linker flags if asked
+if CLANG_ARCH_X86_64
+AM_CFLAGS += -arch x86_64
+AM_LDFLAGS += -arch x86_64
+endif
+
+# Pass arm64 in compiler and linker flags if asked
+# (Note - may be additive to the above)
+if CLANG_ARCH_ARM64
+AM_CFLAGS += -arch arm64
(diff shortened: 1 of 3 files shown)


xnorpx:
@mark4o && @rillian for review?

Should neon be presumed for apple silicone?
Add config to .travis ci for coverage?

silverbacknet:
NEON has been in all Apple silicon since 3GS, so it would be reasonable.

andrewgrant:
Good point RE Neon. I'll add that and resubmit in the next day or so.

rillian:
Can you please also document your rationale in the commit message, or a comment. I assume this is to work around autoconf's built-in cross-compiler detection not supporting Apple's new configuration, but it seems like there are several ways to do that. Why the separate `clang_arch` variables, for example, and why not set the flags directly in configure. That's the sort of thing someone having to maintain the changes later would need to know.

j-schultz:
It's not part of this PR, but I found that the CMake build system is also doing dubious stuff here that pessimizes x64 builds on arm64: https://github.com/xiph/opus/blob/ccaaffa9a3ee427e9401c4dcf6462e378d9a4694/cmake/OpusConfig.cmake#L42

From my understanding, `CMAKE_SYSTEM_PROCESSOR` matches arm64 here, causing the build system to check for NEON availability rather than SSE availability:

```
-- Looking for arm_neon.h
-- Looking for arm_neon.h - not found
```

The resulting binary is still compiled for x64, but without SIMD optimiations.

xnorpx:
@j-schultz so the issue is cross compiling to x64 on a m1/m2 mac for cmake?

j-schultz:
> so the issue is cross compiling to x64 on a m1/m2 mac for cmake?

Correct.

xnorpx:
@j-schultz m1 stuff seems pretty hairy, unlikely I will get fixes in this release.

charlifiiiii:
Is cross-compiling now supposed to work on Mac M1 ?
I tried to cross-compile for Mac Intel (x86_64) from a Mac M1 using this cmd
            ./configure --host=x86_64-apple-macos
and got the following error :
            configure: error: no supported Get CPU Info method, please disable run-time CPU capabilities detection or intrinsics

Any advice ?
Thanks

PR #189, opened 2020-07-12
Fix MinGW problem

z-s-e:
It seems to me MinGW does not properly support _FORTIFY_SOURCE (I get linking errors)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -249,7 +249,7 @@ target_include_directories(
 target_link_libraries(opus PRIVATE ${OPUS_REQUIRED_LIBRARIES})
 target_compile_definitions(opus PRIVATE OPUS_BUILD ENABLE_HARDENING)
 
-if(NOT MSVC)
+if(NOT MSVC AND NOT MINGW)
   target_compile_definitions(opus PRIVATE _FORTIFY_SOURCE=2)
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -120,6 +120,12 @@ cmake_dependent_option(OPUS_STACK_PROTECTOR
                        OFF)
 add_feature_info(OPUS_STACK_PROTECTOR OPUS_STACK_PROTECTOR ${OPUS_STACK_PROTECTOR_HELP_STR})
 
+if(NOT MSVC AND NOT MINGW)
+    set(OPUS_FORTIFY_SOURCE_HELP_STR "add protection against buffer overflows.")
+    option(OPUS_FORTIFY_SOURCE ${OPUS_FORTIFY_SOURCE_HELP_STR} ON)
+    add_feature_info(OPUS_FORTIFY_SOURCE OPUS_FORTIFY_SOURCE ${OPUS_FORTIFY_SOURCE_HELP_STR})
+endif()
+
 if(OPUS_CPU_X86 OR OPUS_CPU_X64)
   set(OPUS_X86_MAY_HAVE_SSE_HELP_STR "does runtime check for SSE1 support.")
   cmake_dependent_option(OPUS_X86_MAY_HAVE_SSE
@@ -249,7 +255,7 @@ target_include_directories(
 target_link_libraries(opus PRIVATE ${OPUS_REQUIRED_LIBRARIES})
 target_compile_definitions(opus PRIVATE OPUS_BUILD ENABLE_HARDENING)
 
-if(NOT MSVC AND NOT MINGW)
+if(OPUS_FORTIFY_SOURCE)
   target_compile_definitions(opus PRIVATE _FORTIFY_SOURCE=2)
 endif()
 


z-s-e:
Ok, found it, [this](https://github.com/msys2/MINGW-packages/issues/5868) is why.

Side note, personally I'm not a fan of forcing anyone to use _FORTIFY_SOURCE like this at all. At least the stack protection is an option, but in my opinion those two things should not be in here at all, they belong in (user provided) toolchain files.
I understand this is done to make default builds safer, maybe an alternative to this would be to just detect when _FORTIFY_SOURCE is not set (externally, e.g. by a toolchain file) and print a warning that using it is recommended...

xnorpx:
Thanks, Ill see if I can make a github action for MinGW builds. I pick up the patch in my next cmake updates.

Regarding your side note cmake has just mostly been a clone to do whatever automake for nix. But I agree we could add it as a Option in CMake.

@jmvalin @mark4o can probably argue why _FORTIFY_SOURC is and should be enabled by default.

z-s-e:
I mean, I know enabling _FORTIFY_SOURCE and stack protection per default is well-intended, and I can see the argument for it. But doing that on the library level (rather than the toolchain or even upstream compiler defaults) is problematic. Having these invasive instrumentation flags enabled as default on the library level have a high chance of breaking builds with less well tested toolchains or other (potentially conflicting) compile options now (I've seen another bug report here about it) and in the future, being a potential constant maintenance burden.

It is a trade-off, to be sure. You'll have to decide if the extra security by default is worth this hassle ;)

mark4o:
Building with _FORTIFY_SOURCE provides extra protection against buffer overflows, with minimal overhead.  It is especially recommended for code that processes untrusted input, such as codecs.

In the autotools build, _FORTIFY_SOURCE is only enabled if it is able to successfully link an executable built with this option: https://github.com/xiph/opus/blob/master/m4/ax_add_fortify_source.m4#L35  It sounds like the CMake build is doing something different.

So if _FORTIFY_SOURCE was broken in MinGW then the autotools build should still work; this option would just not be enabled.  And when _FORTIFY_SOURCE is fixed in MinGW it should automatically get enabled again.

z-s-e:
If you look at the link I provided, to make that work for MinGW (-w64) is to link against another library. I would be kind of annoyed if this suddenly added another runtime dependency on a package, and would really appreciate it if _FORTIFY_SOURCE can be made at least optional, even if you insist it being on per default.

xnorpx:
@z-s-e can you test if autotools check works in MINGW if you have environment setup for it?

We can add similar check as autotools in CMake and simply fail the check if it needs additional libraries linked.

My goal with CMake is do have 1-1 mapping to automake but of course we should make it configurable to the state that it can fit into any bigger build and be customized as such. 

Hopefully we can make everyone somewhat happy and make integration of Opus easy for everyone everywhere.

z-s-e:
Ok, I tried building with autotools and indeed got the same linking error as with cmake. For reference, my setup is Ubuntu 20.04, using the distro packaged MinGW (9.3). With the current master I ran
```
$ ./autogen.h
$ mkdir build; cd build
$ ../configure --host=x86_64-w64-mingw32
$ make -j8
```
Hope I did everything right, never tried MinGW with autotools...

xnorpx:
After spending a little bit on time on this tonight it looks like the only check that autotools is doing is to verify it's not already defined by the system in cflags. And since autotools is failing as well it does not catch this issue. 

We can do it in 2 stages, add Option and then try to write a try_compile check that can catch it. I did not succeed so far.

evpobr:
More info here: msys2/MINGW-packages#5803.

xnorpx:
Based on the discussion in https://github.com/msys2/MINGW-packages/issues/5803 just disabling it by default for MinGW seems fine to me. (we can still add extra option to disable it)

Autotools is still broken however fyi: @mark4o, @rillian

z-s-e:
Well, do you want me to change this PR? Something like this perhaps:
```
if(NOT MSVC AND NOT MINGW)
    set(OPUS_FORTIFY_SOURCE_HELP_STR "add protection against buffer overflows.")
    option(OPUS_FORTIFY_SOURCE ${OPUS_FORTIFY_SOURCE_HELP_STR} ON)
    add_feature_info(OPUS_FORTIFY_SOURCE OPUS_FORTIFY_SOURCE ${OPUS_FORTIFY_SOURCE_HELP_STR})
endif()
...
if(OPUS_FORTIFY_SOURCE)
  target_compile_definitions(opus PRIVATE _FORTIFY_SOURCE=2)
endif()
```
Or do you rather want to do this yourself? I don't mind either way.

xnorpx:
looks good to me, you can update this PR and I will add move it over as patch to gitlab for review. (which is main repo) 

I will also add MinGW to cmake ci at some point.

z-s-e:
Good, thanks

xnorpx:
merge request here: https://gitlab.xiph.org/xiph/opus/-/merge_requests/12 also disabled stack protection by default in mingw builds to make it build clean

xnorpx:
@z-s-e @rillian this has been delivered and can be closed

PR #208, opened 2020-10-31
Initial function for music/speech detection

lstolcman:


lstolcman:
opened by mistake in wrong repo :x

PR #210, opened 2020-11-22
CI - generator script for cmake and github actions

xnorpx:
First iteration of proposal of CI generator script for cmake and github actions.

The script will generate a yaml file for github actions, currently different combinations of build settings and desktop x86|x64 is built.

There is plenty of todo that is not considered in this commit:

- autotools support
- meson support
- gitlab support
- ios, android, arm
- custom scripts (whitespace check etc.)
etc.


xnorpx:
Not sure, why the actions is not showing up here. They might need to be activated. Here is actions running on my fork.

https://github.com/xnorpx/opus/actions/runs/376994457

@rillian @mark4o

xnorpx:
Looks like github actions needs to be activated, they run fine as pr in my fork.

scripts/github_action.py
```diff
@@ -0,0 +1,61 @@
+class GithubActionsTransformer():
+    def __init__(self, name):
+        super(GithubActionsTransformer, self).__init__()
```
rillian, line 3: What does calling `super` do here? AFAICT there's no base class?

scripts/github_action.py
```diff
@@ -0,0 +1,61 @@
...
+        self.actions = {
+            'on': [
+                "push",
+                "pull_request",
+                "pull_request_target",
+            ],
+            "jobs": {
```
rillian, line 10: When I run this, I get the cmake jobs in a different order. Do we need to sort the job keys when writing out `cmake.yml` to avoid churn between python versions, or did you change something and forget to update the output?

.github/workflows/cmake.yml
```diff
@@ -0,0 +1,966 @@
+on:
+- push
+- pull_request
+- pull_request_target
```
rillian, line 4: Since this is a generated file, it would be a good idea to print a comment at the top declaring it so, and pointing people back at `scripts/ci_generator.py` to avoid wasting anyone's time if they're modifying this.

rillian (changes requested):
Thanks for writing this. It looks like a good start, and is certainly an improvement over our current github coverage! :)

My immediate question is whether there's a way to have an action run the generator, instead of having `cmake.yml` checked into version control, with the attendant problems that causes. I think we talked about this before, but I don't remember the answer.

Having seen it all written out, I guess we'll always have to deal with issues specific to each ci and build platform, so it's not actually possible to put every definition in one place. Still, I think it's worthwhile having the script to write out the matricies (RIP travis-ci) and avoid duplicating as much as we can.

If you can follow up on my questions, I'm ok with it landing with the current design. In the meantime I'll see if I can get it generating `.gitlab-ci.yml` as well.

scripts/github_action.py
xnorpx: good catch, I had a base class that felt redundant but forgot to clean up

scripts/github_action.py
xnorpx: Good idea, let me add sort

.github/workflows/cmake.yml
xnorpx: yes, had it on my todo but got stuck in yaml land, will iron this out in some way.

xnorpx:
@rillian I think the autogenerating as CI step is probably just gonna add complexity, maybe easy todo in one ci system but hard in another etc. However, maybe we can add a actual CI step that runs the generation on the CI server and do a diff on the generated files to make sure they are updated and commited, this should work across CI systems (assuming they have python) and can act as a sanity check.

xnorpx:
I have some work to catch up to after holiday so will work on this end of week hopefully.

scripts/github_action.py
xnorpx: fixed

.github/workflows/cmake.yml
xnorpx: fixed.

scripts/github_action.py
xnorpx: fixed.

xnorpx:
I will look into verification of the generation and I have some other todos I would like to polish i.e. no merge yet.

PR #211, opened 2020-11-26
Meson: Fix doc build when opus is a subproject

xclaesse:
meson.source_root() and meson.build_root() have been deprecated in
latest Meson release because they are a trap. They point to the root of
parent project instead of root of current subproject. Meson 0.56.0 added
meson.project_source/build_root() but it is just as easy to use
meson.current_source/build_dir() in the root meson.build file and avoids
bumping required meson version.

### DIFF
--- a/doc/meson.build
+++ b/doc/meson.build
@@ -1,6 +1,3 @@
-top_srcdir = meson.source_root()
-top_builddir = meson.build_root()
-
 have_dot = find_program('dot', required: false).found()
 
 doxyfile_conf = configuration_data()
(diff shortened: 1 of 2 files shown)


dabrain34:
This needs at least 0.56.0 ?

xclaesse:
No, I,m using meson.current_source_dir() that has been in meson for a while instead of the new meson.project_source_root() that is new in 0.56.

dabrain34:
sorry I read uncorrectly as I was still on what we discussed earlier.

tp-m:
Thanks for fixing this.

Fix looks good to me.

rillian:
Thanks @xclaesse for the fix, and to @tp-m and @dabrain34 for review. Merged upstream in d2f6805c3f36718a742c33c56eb945c2e10646ef.

PR #219, opened 2021-02-12
Sending refresh DTX packets every 400 ms independently of the encoded…

jvicentep:
… frame size.

This pull request makes sure that refresh DTX packets are sent every 400 ms. When encoding with frame sizes different from 20 ms and using Silk (for example by forcing a low bitrate) the distance between refresh DTX packets was different from 400 ms. Specifically, for frames length equal to 60 ms, the packets were at 1.2 seconds distance.

As an example, the attached image compares a DTX region before and after the change proposed in this pull request. We can see that the upper image did not receive a refresh DTX packet for 1.2 second and, therefore, kept a high level of comfort noise for longer time. The bottom spectrogram was created with the proposed changes. 

example_DTX_packet_duration

### DIFF
--- a/src/opus_encoder.c
+++ b/src/opus_encoder.c
@@ -112,7 +112,7 @@ struct OpusEncoder {
     opus_val16   delay_buffer[MAX_ENCODER_BUFFER*2];
 #ifndef DISABLE_FLOAT_API
     int          detected_bandwidth;
-    int          nb_no_activity_frames;
+    int          nb_no_activity_ms_Q1;
     opus_val32   peak_signal_energy;
 #endif
     int          nonfinal_frame; /* current frame is not the final in a packet */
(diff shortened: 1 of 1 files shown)


src/opus_encoder.c
```diff
@@ -893,24 +893,28 @@ static opus_val32 compute_frame_energy(const opus_val16 *pcm, int frame_size, in
...
    {
-      /* The number of consecutive DTX frames should be within the allowed bounds */
-      (*nb_no_activity_frames)++;
-      if (*nb_no_activity_frames > NB_SPEECH_FRAMES_BEFORE_DTX)
+      /* The number of consecutive DTX frames should be within the allowed bounds. 
+      Note that the allowed bound is defined in the Silk headers and assumes 20 ms
+      frames. As this function can be called with any frame length, a conversion to
+      miliseconds is done before the comparisons. */
+      (*nb_no_activity_ms_Q2) += frame_size_ms_Q2;
+      if (*nb_no_activity_ms_Q2 > NB_SPEECH_FRAMES_BEFORE_DTX*20*2)
```
xnorpx, line 908: I don't get the *2

jvicentep: that counter, nb_no_activity_ms_Q2, is using a Q2 precission as we have to cover the case of 2.5 ms frames and we are using integers. So the *2 is needed for converting the NB_SPEECH_FRAMES_BEFORE_DTX to miliseconds in Q2.

xnorpx: I see makes sense, I would probably add new defines with MS and Q2 in the names but might be overengineering it.

xnorpx: also for old time sakes, what about test ;)

jvicentep: I realize now that it is Q1, I will update the pull request. Yes, tests ... the ones that did not exist :) I have not looked yet how easy it is to add new ones.

jmvalin:
Just curious, does that fix or improve a particular behaviour. I was wondering whether 400 ms is still appropriate for longer packets or if we might want to make the delay larger -- on the assumption that bandwidth is tight. Just asking.

jvicentep:
This change makes the Opus behavior a bit more consistent and fits better with the RFC specifications (https://tools.ietf.org/html/rfc6716#section-2.1.9). 

This bug is not always happening and it depends on the configuration for Opus. For example, if the floating point code is disabled, this bug is not present. So, this change also makes the code more consistent over different Opus configurations. 

Quality wise, I do not think this change has a large impact except for some corner cases like the one that I showed in the figure.

felicialim:
Merged in https://gitlab.xiph.org/xiph/opus/-/commit/16286a25fdd865c66a837a73b65fbaa7b25bf484

xnorpx:
this has been merged and can be closed @mark4o

PR #220, opened 2021-02-22
Tiny speed up for floating-point targets

6ws:


xnorpx:
you are replacing qi = round(f) with qi = 1?

xnorpx:
Just to clarify, the 0.5f just indicates that 0.5 is single precision and is not the same as the variable f.

6ws:
> Just to clarify, the 0.5f just indicates that 0.5 is single precision and is not the same as the variable f.

Yep, you're correct so I'll close PR. My bad.

Issue #256, opened 2022-07-18
build failed when disable AVX on Windows with MSVC clang-cl compiler


wangyoucao577:
Hi,        

I'm trying to disable `AVX` on windows compilation to compatible with old CPUs that no `AVX` instructions support. I uses  `clang-cl` compiler(clang 12) that installed in latest Visual Studio 2022, but the build is failed. Here's the reproduction and error logs.     

```pwsh
> mkdir -p build && cd build 
> cmake .. -GNinja -DCMAKE_C_COMPILER=clang-cl -DOPUS_X86_MAY_HAVE_AVX=OFF
> ninja
[145/149] Building C object CMakeFiles\opus.dir\silk\x86\VQ_WMat_EC_sse4_1.c.obj
FAILED: CMakeFiles/opus.dir/silk/x86/VQ_WMat_EC_sse4_1.c.obj
C:\PROGRA~1\MICROS~3\2022\COMMUN~1\VC\Tools\Llvm\bin\clang-cl.exe  /nologo -DENABLE_HARDENING -DHAVE_CONFIG_H -DHAVE_LRINT -DHAVE_LRINTF -DOPUS_BUILD -DOPUS_HAVE_RTCD -DOPUS_X86_MAY_HAVE_SSE -DOPUS_X86_MAY_HAVE_SSE2 -DOPUS_X86_MAY_HAVE_SSE4_1 -DVAR_ARRAYS -D_CRT_SECURE_NO_WARNINGS -IE:\opus\include -IE:\opus\build -IE:\opus -IE:\opus\celt -IE:\opus\silk -IE:\opus\silk\float /DWIN32 /D_WINDOWS /W3 /MDd /Zi /Ob0 /Od /RTC1 /GS /showIncludes /FoCMakeFiles\opus.dir\silk\x86\VQ_WMat_EC_sse4_1.c.obj /FdCMakeFiles\opus.dir\opus.pdb -c -- E:\opus\silk\x86\VQ_WMat_EC_sse4_1.c
E:\opus\silk\x86\VQ_WMat_EC_sse4_1.c(88,26): error: always_inline function '_mm_cvtepi8_epi32' requires target feature 'sse4.1', but would be inlined into function 'silk_VQ_WMat_EC_sse4_1' that is compiled without support for 'sse4.1'
        v_cb_row_31_Q7 = OP_CVTEPI8_EPI32_M32( &cb_row_Q7[ 1 ] );
                         ^
[...]

PR #257, opened 2022-07-23
use -mavx clang style options to enable SIMD for clang-cl

wangyoucao577:
- Only use `/arch:AVX` MSVC style options to enable SIMD when strictly MSVC cl;        
- Use `-mavx` gcc/clang style options to enable SIMD when possible, for example clang-cl installed in MSVC;      
- Enable `C11` for `clang-cl`;        
- Enable cpuinfo for `clang-cl`;

Closes https://github.com/xiph/opus/issues/256


CMakeLists.txt
```diff
@@ -389,13 +389,13 @@ if(NOT OPUS_DISABLE_INTRINSICS)
     if(OPUS_X86_MAY_HAVE_SSE)
       add_sources_group(opus celt ${celt_sources_sse})
       target_compile_definitions(opus PRIVATE OPUS_X86_MAY_HAVE_SSE)
-      if(NOT MSVC)
+      if(NOT (MSVC AND CMAKE_C_COMPILER_ID STREQUAL "MSVC"))
```
xnorpx, line 392: Maybe easier to make a variable for the case when using MSVC compiler frontend and not.

xnorpx:
@wangyoucao577 can you add the same check for all other compiler flags regarding clang-cl frontend. (fast-math etc)

CMakeLists.txt
wangyoucao577: Good point! Updated.

wangyoucao577:
> @wangyoucao577 can you add the same check for all other compiler flags regarding clang-cl frontend. (fast-math etc)

I checked all `MSVC` related options in current cmake, unfortunately `-ffast-math` can't work on `clang-cl`, it must be `/fp:fast`. Supported options(c11, cpuinfo) has been updated.

jmvalin:
What's the status of this PR?

xnorpx:
Master branch is deleted and all pr's against master is closed. If this change is still relevant please reopen and repoint your PR to main branch.

PR #277, opened 2023-04-26
meson: Install pkgconfig's pc file in correct path on FreeBSD

diizzyy:
The builtin function automatically sets correct install path but since we're working around it we need to make sure manually


xclaesse:
I would expect Meson generator to install in the right place on all systems, if not that should be fixed in Meson. Better use Meson generator: https://github.com/xiph/opus/pull/286.

PR #284, opened 2023-06-17
Add more configurations for Github Action and remove legacy visual studio solution

xnorpx:
Remove Visual Studio solution which be replaced by Meson and CMake
Remove travis and appveyor CI which is replaced with Github Action CI


xnorpx:
moved to gitlab for merge.

PR #286, opened 2023-06-20
meson: Use pkgconfig generator

xclaesse:
It is much less error prone because Meson can detect dependencies automatically. As bonus Meson will also generate the opus-uninstalled.pc file.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -586,29 +586,11 @@ if not opt_tests.disabled()
   subdir('tests')
 endif
 
-# pkg-config files (not using pkg module so we can use the existing .pc.in file)
-pkgconf = configuration_data()
-
-pkgconf.set('prefix', join_paths(get_option('prefix')))
-pkgconf.set('exec_prefix', '${prefix}')
-pkgconf.set('libdir', '${prefix}/@0@'.format(get_option('libdir')))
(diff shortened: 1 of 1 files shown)


luzpaz:
Is there an ETA on this ?

diizzyy:
It would indeed be nice if we could have this merged

heitbaum:
I did a quick test on this and the .pc file is operationally the same as before.
```
diff -Nu opus.pc opus.pc.286
--- opus.pc     2024-03-14 13:24:42.326119773 +0000
+++ opus.pc.286 2024-03-14 13:30:32.419416360 +0000
@@ -1,16 +1,9 @@
-# Opus codec reference implementation pkg-config file
-
 prefix=/usr
-exec_prefix=${prefix}
-libdir=${prefix}/lib
 includedir=${prefix}/include
+libdir=${prefix}/lib
 
-Name: Opus
+Name: opus
 Description: Opus IETF audio codec (floating-point build)
-URL: https://opus-codec.org/
 Version: 1.4-git-a8e4ebb
-Requires:
-Conflicts:
-Libs: -L${libdir} -lopus
-Libs.private: -lm
+Libs: -L${libdir} -lopus -lm
 Cflags: -I${includedir}/opus
```

jmvalin:
merged

PR #296, opened 2023-08-21
meson fix comparison

heitbaum:
cc.get_define returns str (not bool)

Fixes:
Fetching value of define "__APPLE__" :  
../meson.build:316:12: ERROR: Object <[StringHolder] holds [str]: ''> of type str does not support the `bool()` operator.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -313,7 +313,7 @@ if not opt_asm.disabled()
         endif
       endif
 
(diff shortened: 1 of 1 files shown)


mklingb:
LGTM

jmvalin:
merged

PR #307, opened 2023-12-29
dump_modes: add missing file to build

rasky:
The file is required to build dump_modes correctly, because it defines `celt_assert`.


jmvalin:
Merged, thanks.

PR #310, opened 2024-01-22
Provide 5th order HOA channel map 3 mixing and demixing

chris-hld:
Starting implementation and discussion of higher order Ambisonic optimizations.


jmvalin:
Just curious, did you sync up with the folks at Google who contributed the original ambisonics support? There doesn't seem to be anything too risky in there, but best to have someone reviewing who knows what they're doing.

jmvalin:
Merged in main branch (master is deprecated).
Note that I had to rewrap the lines for the matrix and fixed the Opus projection test.

chris-hld:
Great!

> Just curious, did you sync up with the folks at Google who contributed the original ambisonics support? 

We had a chat end of last year, but the person which worked the most on it back then was out of office.
I'll ping them again regarding further optimizations, I'll CC you if you're interested.

The 5th order was mostly important to reference my own projects, here is the 4th order so that there is now support for all orders up to 5 https://github.com/xiph/opus/pull/314

PR #314, opened 2024-02-15
Provide 4th order HOA map 3 mixing and demixing

chris-hld:
Enables all HOA orders up to 5.


chris-hld:
The two extra channels in mixing / demixing for 5th order were not passed through yet. 
Now consistent between orders.
Also, casting to int now, instead of rounding, therefore the slight difference in some of the numbers.

jmvalin:
Thanks. Merged your changes to main.

PR #339, opened 2024-04-24
Fix opus_demo lossgen when using weights file

jim-signal:
When trying to test `opus_demo` with both `USE_WEIGHTS_FILE` and `ENABLE_LOSSGEN`, it does not work. The "loss generator" data is invalid and therefore does not apply the loss.

This PR adds the lossgen data to the lpcnet weights file and loads it when running the `opus_demo`, when needed.

### DIFF
--- a/dnn/write_lpcnet_weights.c
+++ b/dnn/write_lpcnet_weights.c
@@ -50,6 +50,9 @@
 #include "lace_data.c"
 #include "nolace_data.c"
 #endif
+#ifdef ENABLE_LOSSGEN
+#include "lossgen_data.c"
+#endif
 
 void write_weights(const WeightArray *list, FILE *fout)
 {
@@ -91,6 +94,9 @@ int main(void)
 #ifndef DISABLE_NOLACE
   write_weights(nolacelayers_arrays, fout);
 #endif
+#endif
+#ifdef ENABLE_LOSSGEN
+  write_weights(lossgen_arrays, fout);
 #endif
   fclose(fout);
   return 0;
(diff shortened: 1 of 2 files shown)


jmvalin:
Actually, on this one I think the best fix may be to always include the lossgen weights in the build rather than the blob. After all, the weights aren't in libopus and opus_demo isn't meant to be shipped anyway. Makes sense?

jim-signal:
Yes, I agree. I had tried that before but it wasn't working. However, I find that if I add
```
#undef USE_WEIGHTS_FILE
```
to the `lossgen_data.h`, that works. The `lossgen_init()` function in `lossgen.c` also needs the `USE_WEIGHTS_FILE` block adjusted.

I am not sure where lossgen_data.* is generated to adjust that or if you'd prefer to also remove all references to `USE_WEIGHTS_FILE` in lossgen_data.c.

jmvalin:
@janpbuethe Any idea if there's an easy way to not have the #ifdef USE_WEIGHTS_FILE when dumping the weights? The alternative would be to have lossgen.c #include "lossgen_data.c" after an #undef USE_WEIGHTS_FILE

janpbuethe:
CWriter in the weight-exchange library has an argument enable_binary_blob which defaults to True. Setting it to False should (in theory) solve the problem.

jmvalin:
@jim-signal  can you give the exp_lossgen_fix1 branch a try and see if it fixes the problem?

jim-signal:
Yes, that works. If I turn on `USE_WEIGHTS_FILE`:
- `opus_demo` is about 670KB and properly does not include the weight data
- `./dump_weights_blob` yields a ~3.3MB weights_blob.bin
- Running `opus_demo` with `sim_loss` and with/without DRED results in expected results

Thanks! Feel free to close this and merge your own patch.

jmvalin:
@jim-signal So I merged a more complete patch to main. Let me know if there's any issue with it.

PR #347, opened 2024-05-11
Add sse4.1 flag when using clang-cl on MSVC targets

DoumanAsh:
Without flag specified you shall get errors from clang about sse4.1 feature missing:
```
 error: '__builtin_ia32_pblendw128' needs target feature sse4.1
```


jmvalin:
merged

PR #350, opened 2024-06-12
Fix SIMD fallback code when AVX2 is not available 

j-schultz:
Same as https://github.com/xiph/rnnoise/pull/228

Currently the code first typedefs a fallback type based on __m256i and only on the following line it actually defines __m256i as the emulated type. This needs to be the other way around! This fixes building opus on emscripten with vectorization enabled.


jmvalin:
Should now be fixed by https://github.com/xiph/opus/commit/b4909e1

Issue #361, opened 2024-07-29
compiler AVX2 intrinsics check is failing in autotools.


sezero:
This is on i686-linux with gcc-14. Relevant parts of config.log:

```
configure:15283: checking if compiler supports AVX2 intrinsics with -mavx -mfma -mavx2
configure:15312: gcc14 -o conftest -g -O2 -mavx -mfma -mavx2   conftest.c  >&5
conftest.c: In function 'main':
conftest.c:40:53: error: passing argument 1 of '_mm_loadu_si128' from incompatible pointer type [-Wincompatible-pointer-types]
   40 |               _mm256_cvtepi16_epi32(_mm_loadu_si128(utest));
      |                                                     ^~~~~
      |                                                     |
      |                                                     unsigned char *
In file included from /opt/gcc14/lib/gcc/i686-pc-linux-gnu/14.1.1/include/xmmintrin.h:1322,
                 from /opt/gcc14/lib/gcc/i686-pc-linux-gnu/14.1.1/include/immintrin.h:31,
                 from conftest.c:25:
/opt/gcc14/lib/gcc/i686-pc-linux-gnu/14.1.1/include/emmintrin.h:704:35: note: expected 'const __m128i_u *' but argument is of type 'unsigned char *'
  704 | _mm_loadu_si128 (__m128i_u const *__P)
      |                  ~~~~~~~~~~~~~~~~~^~~
configure:15312: $? = 1
configure: failed program was:
| /* confdefs.h */
| #define PACKAGE_NAME "opus"
| #define PACKAGE_TARNAME "opus"
| #define PACKAGE_VERSION "1.5.2-12-gfc96bb9"
| #define PACKAGE_STRING "opus 1.5.2-12-gfc96bb9"
| #define PACKAGE_BUGREPORT "opus@xiph.org"
| #define PACKAGE_URL ""
| #define HAVE_STDIO_H 1
| #define HAVE_STDLIB_H 1
| #define HAVE_STRING_H 1
| #define HAVE_INTTYPES_H 1
| #define HAVE_STDINT_H 1
| #define HAVE_STRINGS_H 1
| #define HAVE_SYS_STAT_H 1
| #define HAVE_SYS_TYPES_H 1
| #define HAVE_UNISTD_H 1
| #define STDC_HEADERS 1
| #define HAVE_DLFCN_H 1
| #define LT_OBJDIR ".libs/"
| #define OPUS_BUILD /**/
| #define restrict __restrict
[...]

PR #362, opened 2024-07-30
configure.ac: cast utest to __m128i_u* in avx2 check for _mm_loadu_si128

sezero:
Fixes avx2 check.
Fixes: https://github.com/xiph/opus/issues/361.


configure.ac
```diff
@@ -697,9 +697,10 @@ AS_IF([test x"$enable_intrinsics" = x"yes"],[
              __m256i mtest2;
              mtest = _mm256_set1_ps((float)time(NULL));
              mtest = _mm256_fmadd_ps(mtest, mtest, mtest);
-             mtest1 = _mm256_set_m128i(_mm_loadu_si64(utest), _mm_loadu_si64(utest));
+             mtest1 = _mm256_set_m128i(
+              _mm_loadl_epi64((__m128i_u *)utest), _mm_loadl_epi64((__m128i_u *)utest));
```
jmvalin, line 701: As far as I can tell, __m128i_u is just an internal gcc/clang thing and not something we can rely on unfortunately.

sezero: Solution?

PR #363, opened 2024-08-18
Add Arm RTCD for OpenBSD

brad0:
Also fix FreeBSD CMake support on ARM.

### DIFF
--- a/celt/arm/armcpu.c
+++ b/celt/arm/armcpu.c
@@ -191,7 +191,7 @@ static opus_uint32 opus_cpu_capabilities(void)
   return flags;
 }
 
-#elif defined(__FreeBSD__)
+#elif defined(HAVE_ELF_AUX_INFO)
 #include <sys/auxv.h>
 
 static opus_uint32 opus_cpu_capabilities(void)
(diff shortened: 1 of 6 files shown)


brad0:
cc @jmvalin

brad0:
@jmvalin Ping.

celt/arm/armcpu.c
```diff
@@ -191,7 +191,7 @@ static opus_uint32 opus_cpu_capabilities(void)
   return flags;
 }
 
-#elif defined(__FreeBSD__)
+#elif defined(HAVE_ELF_AUX_INFO)
```
jmvalin, line 194: I'm not sure I understand why you need to change the __FreeBSD__ case to add OpenBSD. Not against your OpenBSD patch but would like to make sure we don't break FreeBSD. More specifically, your HAVE_ELF_AUX_INFO is added to only two out of the four build systems we support.

brad0:
The meson build system seems to be broken.

```
arm64$ meson setup build
The Meson build system
Version: 1.8.2
Source dir: /home/brad/tmp/opus
Build dir: /home/brad/tmp/opus/build
Build type: native build

meson.build:2:11: ERROR: Command `/home/brad/tmp/opus/meson/get-version.py --package-version` failed with status 255.

A full log can be found at /home/brad/tmp/opus/build/meson-logs/meson-log.txt
```

celt/arm/armcpu.c
brad0: Not every release of FreeBSD nor OpenBSD has the elf_aux_info() API. This is how I've done things with pretty much every other project I've added OpenBSD or FreeBSD/OpenBSD support. I have now added and fixed the CMake support.

brad0:
@jmvalin Ping.

PR #366, opened 2024-09-17
Fix _mm_loadu_si64 not available in GCC8

razvanphp:
This preserves BC while allowing compilation on older systems like Debian 10.

Inspired from https://github.com/ggerganov/llama.cpp/pull/1154

### DIFF
--- a/silk/x86/NSQ_del_dec_avx2.c
+++ b/silk/x86/NSQ_del_dec_avx2.c
@@ -956,7 +956,7 @@ static OPUS_INLINE void silk_nsq_del_dec_scale_states_avx2(
     inv_gain_Q26 = silk_sar_round_32(inv_gain_Q31, 5);
     for (i = 0; i < psEncC->subfr_length; i+=4)
     {
-        __m256i x = _mm256_cvtepi16_epi64(_mm_loadu_si64(&x16[i]));
+        __m256i x = _mm256_cvtepi16_epi64(_mm_loadl_epi64(&x16[i]));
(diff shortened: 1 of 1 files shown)


silk/x86/NSQ_del_dec_avx2.c
```diff
@@ -956,7 +956,7 @@ static OPUS_INLINE void silk_nsq_del_dec_scale_states_avx2(
     inv_gain_Q26 = silk_sar_round_32(inv_gain_Q31, 5);
     for (i = 0; i < psEncC->subfr_length; i+=4)
     {
-        __m256i x = _mm256_cvtepi16_epi64(_mm_loadu_si64(&x16[i]));
+        __m256i x = _mm256_cvtepi16_epi64(_mm_loadl_epi64(&x16[i]));
```
jmvalin, line 959: _mm_loadl_epi64() takes a __m128i *, which means the PR would break alignment rules. I suspect there should still be a workaround...

sezero: Both _mm_loadl_epi64 and _mm_loadu_epi64 take `void const *`
From gcc-14:
```c
extern __inline __m128i __attribute__((__gnu_inline__, __always_inline__, __artificial__))
_mm_loadu_si64 (void const *__P)
{
  return _mm_loadl_epi64 ((__m128i_u *)__P);
}
/* */
extern __inline __m128i
__attribute__ ((__gnu_inline__, __always_inline__, __artificial__))
_mm_loadu_epi64 (void const *__P)
{
  return (__m128i) (*(__v2di_u *) __P);
}
```

Maybe just adding an `(__m128i_u *)` cast would be enough??

PR #369, opened 2024-10-07
Support OPUS_SET_COMPLEXITY() for multistream decoder

dddaniel:
OPUS_SET_COMPLEXITY() is already implemented in the decoder for controlling deep PLC since commit
8d43b185b2bb44841cb9b13e6dc1f4f30da0db87

This commit makes it also available for the mulitstream decoder.


eshaz:
@jmvalin I tried this out with the multistream decoder, and this change is required to get the decoder to enable the new LACE and NoLACE decoding features. Any news on when this can be merged?

PR #370, opened 2024-10-17
Fix warning for static vs non-static declaration/definitions

dimitry-:
In include/opus_custom.h these are declared as OPUS_CUSTOM_EXPORT_STATIC


jmvalin:
We *want* these functions to be static when not building with custom modes enabled. I'm not sure where the warning comes from, but it would have a different cause.

PR #373, opened 2024-10-25
Add OPUS_SET_MIN_BANDWIDTH to set minInternalSampleRate

nguditi:
… (which is using NARROWBAND)

- Implement OPUS_SET_MIN_BANDWIDTH to OPUS_BANDWIDTH_WIDEBAND to allow configuration of minInternalSampleRate in Silk mode during encoding.
- This enables additional options for decoding, supporting both LACE and NOLACE modes, which require Wideband audio.

### DIFF
--- a/include/opus_defines.h
+++ b/include/opus_defines.h
@@ -133,6 +133,8 @@ extern "C" {
 #define OPUS_GET_BITRATE_REQUEST             4003
 #define OPUS_SET_MAX_BANDWIDTH_REQUEST       4004
 #define OPUS_GET_MAX_BANDWIDTH_REQUEST       4005
+#define OPUS_SET_MIN_BANDWIDTH_REQUEST       40050
+#define OPUS_GET_MIN_BANDWIDTH_REQUEST       40051
 #define OPUS_SET_VBR_REQUEST                 4006
 #define OPUS_GET_VBR_REQUEST                 4007
 #define OPUS_SET_BANDWIDTH_REQUEST           4008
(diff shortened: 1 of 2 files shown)


jmvalin:
Can you explain a bit more in what context you would be using it that doesn't currently work?

nguditi:
> Can you explain a bit more in what context you would be using it that doesn't currently work?

Hi jmvalin. 

When audio passes through the encoder, it’s limited to a 4 kHz frequency, which reduces the quality to narrowband. This also prevents the LACE/NOLACE features from working, as they require wideband, which needs a minimum of 8 kHz. To resolve this and enable LACE/NOLACE, you need to set the encoder’s bandwidth to wideband by calling the following command:

opus_encoder_ctl(encoder, OPUS_SET_BANDWIDTH(OPUS_BANDWIDTH_WIDEBAND));

But this command also limits the upper bound and sets the bandwidth strictly to wideband. However, I want to avoid restricting the upper bound for higher-quality audio. I simply want to set the lower bound to enable LACE/NOLACE for low-quality audio while keeping the original algorithm for medium and high-quality audio.

I hope you understand the situation I'm facing.

Thank you

nguditi:
Let me tell you a story.

I want to test the Low-Bitrate Speech Quality Enhancement feature of Opus v1.5. My initial thought was to simply run the provided docs/trivial_example.c with a bitrate limit down to about 3 kbps for 16-bit audio with 1 channel, or 6 kbps for 32-bit audio with 1 channel, along with `opus_decoder_ctl(decoder, OPUS_SET_COMPLEXITY(7))`. However, the output file quality remains poor, as if the complexity setting wasn’t applied to the decoder. I looked into it further and realized this feature only works with wideband audio quality. I then downloaded an audio file from the Opus demo site, analyzed it in Audacity, and saw that their 32-bit, 1-channel audio was at 8 kHz with a bitrate of 6 kbps. How did they manage that encoding? It turns out that you need to use the configuration to set a fixed bandwidth for the encoder. 

..and finally, the output sounds as good as in the Opus demo.

PR #379, opened 2024-12-05
Arm: Speed up FLOAT2INT16 conversion with Neon

agosdahu:
Using Neon for float to int conversion and introducing platform specific function for converting an array of float values to int16. Also adding appropriate unit test.

### DIFF
--- a/celt/arm/celt_neon_intr.c
+++ b/celt/arm/celt_neon_intr.c
@@ -1,4 +1,5 @@
 /* Copyright (c) 2014-2015 Xiph.Org Foundation
+   Copyright (c) 2024 Arm Limited
    Written by Viswanath Puttagunta */
 /**
    @file celt_neon_intr.c
@@ -35,7 +36,57 @@
 #endif
 
 #include <arm_neon.h>
+#include "../float_cast.h"
+#include "../mathops.h"
 #include "../pitch.h"
+#if defined(OPUS_CHECK_ASM)
+#include <stdlib.h>
+#endif
+
+#if !defined(DISABLE_FLOAT_API) && defined(OPUS_ARM_MAY_HAVE_NEON_INTR)
+
+void celt_float2int16_neon(const float * OPUS_RESTRICT in, short * OPUS_RESTRICT out, int cnt)
+{
+   int i = 0;
+
+#if defined(__ARM_NEON)
+   const int BLOCK_SIZE = 16;
+   const int blockedSize = cnt / BLOCK_SIZE * BLOCK_SIZE;
+
+   for (; i < blockedSize; i += BLOCK_SIZE)
+   {
+      float32x4_t orig_a = vld1q_f32(&in[i +  0]);
+      float32x4_t orig_b = vld1q_f32(&in[i +  4]);
+      float32x4_t orig_c = vld1q_f32(&in[i +  8]);
+      float32x4_t orig_d = vld1q_f32(&in[i + 12]);
+
+      int16x4_t asShort_a = vqmovn_s32(vroundf(vmulq_n_f32(orig_a, CELT_SIG_SCALE)));
+      int16x4_t asShort_b = vqmovn_s32(vroundf(vmulq_n_f32(orig_b, CELT_SIG_SCALE)));
+      int16x4_t asShort_c = vqmovn_s32(vroundf(vmulq_n_f32(orig_c, CELT_SIG_SCALE)));
+      int16x4_t asShort_d = vqmovn_s32(vroundf(vmulq_n_f32(orig_d, CELT_SIG_SCALE)));
+
+      vst1_s16(&out[i +  0], asShort_a);
+      vst1_s16(&out[i +  4], asShort_b);
+      vst1_s16(&out[i +  8], asShort_c);
+      vst1_s16(&out[i + 12], asShort_d);
+# if defined(OPUS_CHECK_ASM)
+      short out_c[BLOCK_SIZE];
+      int j;
+      for(j = 0; j < BLOCK_SIZE; j++)
+      {
+         out_c[j] = FLOAT2INT16(in[i + j]);
+         celt_assert(abs((out_c[j] - out[i + j])) <= 1);
+      }
+# endif
+   }
+#endif
+
+   for (; i < cnt; i++)
+   {
+      out[i] = FLOAT2INT16(in[i]);
+   }
+}
+#endif
 
 #if defined(FIXED_POINT)
 #include <string.h>
--- a/celt/arm/arm_celt_map.c
+++ b/celt/arm/arm_celt_map.c
@@ -1,5 +1,6 @@
 /* Copyright (c) 2010 Xiph.Org Foundation
- * Copyright (c) 2013 Parrot */
+ * Copyright (c) 2013 Parrot
+ * Copyright (c) 2024 Arm Limited */
 /*
    Redistribution and use in source and binary forms, with or without
    modification, are permitted provided that the following conditions
@@ -29,11 +30,24 @@
 #include "config.h"
 #endif
 
-#include "pitch.h"
 #include "kiss_fft.h"
+#include "mathops.h"
 #include "mdct.h"
+#include "pitch.h"
 
 #if defined(OPUS_HAVE_RTCD)
+
+# if !defined(DISABLE_FLOAT_API)
+#  if defined(OPUS_ARM_MAY_HAVE_NEON_INTR) && !defined(OPUS_ARM_PRESUME_NEON_INTR)
+void (*const CELT_FLOAT2INT16_IMPL[OPUS_ARCHMASK+1])(const float * OPUS_RESTRICT in, short * OPUS_RESTRICT out, int cnt) = {
+  celt_float2int16_c,   /* ARMv4 */
+  celt_float2int16_c,   /* EDSP */
+  celt_float2int16_c,   /* Media */
+  celt_float2int16_neon,/* NEON */
+  celt_float2int16_neon /* DOTPROD */
+};
+#  endif
+# endif
 
 # if defined(OPUS_ARM_MAY_HAVE_NEON_INTR) && !defined(OPUS_ARM_PRESUME_NEON_INTR)
 opus_val32 (*const CELT_INNER_PROD_IMPL[OPUS_ARCHMASK+1])(const opus_val16 *x, const opus_val16 *y, int N) = {
--- a/celt/arm/mathops_arm.h
+++ b/celt/arm/mathops_arm.h
@@ -1,0 +1,65 @@
+/* Copyright (c) 2024 Arm Limited */
+/*
+   Redistribution and use in source and binary forms, with or without
+   modification, are permitted provided that the following conditions
+   are met:
+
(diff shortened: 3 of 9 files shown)


jmvalin:
Haven't yet had time to look at all the details, but generally looks like a good thing to add. Do you have any figures for the actual speedup this change provides?

agosdahu:
Running the same sequence of calling opus_demo on all twelve .bit files for both mono and stereo setup @48k sampling rate on a single Cortex A55 core @1.8Ghz gave us 14% of performance uplift when measured via linux perf tool.

Since testvector09.bit and testvector10.bit are the largest test inputs, we also ran a sequence of a 100-100 consecutive stereo decodings for each file @48k sampling rate.
Measurements for this run gave us 11% of performance uplift in the same test environment as above.

celt/arm/celt_neon_intr.c
```diff
@@ -35,8 +36,46 @@
...
+      vst1_s16(&out[i +  8], asShort_c);
+      vst1_s16(&out[i + 12], asShort_d);
+   }
+#endif
+
+   for (; i < cnt; i++)
+   {
+      out[i] = FLOAT2INT16(in[i]);
+   }
+}
```
jmvalin, line 76: Would be good to add an OPUS_CHECK_ASM block to verify that the results match the C code. You can grep for OPUS_CHECK_ASM to see how it's done in other parts of the code

jmvalin:
So the patch looks good. See my comment about OPUS_CHECK_ASM. Otherwise, it would be good if you could run the automated tests to see that nothing breaks. 
To do that, you'll want to go to the tests directory and do:
./opus_build_test.sh <tarball> <count> <opus_testvectors> <opus_newvectors>
where:
<tarball> is the output of a "make dist" with your code
<count> is the number of random builds to run (e.g. 1000)
<opus_testvectors> points to the old testvector directory (https://archive.mozilla.org/pub/opus/opus_testvectors-rfc8251.tar.gz)
<opus_newvectors> points to the new testvector directory (https://opus-codec.org/docs/opus_testvectors-rfc8251.tar.gz)

To test on ARM, you'll just need to change this line from the random_config.sh script:
arch=`echo -e "\n-march=core2\n-march=sandybridge\n-march=broadwell" | shuf -n1`
to use -march options for your compiler. 

If you run into issues with this test, I can run it myself in a few days, but the fastest ARM setup I have is a RPi5 so I won't be able to run that many tests.

jmvalin:
Strange. Assuming the new and old testvectors weren't reversed, can you try running just the testvectors manually? You can do so with:
./tests/run_vectors.sh <path> <testvectors> 48000
where:
<path> is the directory where opus_demo and opus_compare are located
<testvectors> is the directory that has the new testvectors (unless you built with --disable-rfc8251 in which case it would be the old testvectors)

agosdahu:
I ran the` run_vectors.sh` script for both old and new vector sets and both returned with `"All tests have passed successfully"`.
Side note: when inspecting the log_<mono|stereo>[2] files, there were often cases where interim results had `"Test vector FAILS"` and `Internal weighted error` was a relatively large number (compared to the others)

I reconfigured, however, my build with `--disable-rfc8251` and every `run_vectors.sh` run failed (with old and new vector set alike)

jmvalin:
Oh, I think I gave you the wrong URL for the old testvectors. Try this: https://www.opus-codec.org/docs/opus_testvectors.tar.gz
Sorry about that

celt/arm/celt_neon_intr.c
agosdahu: I would be happy to add this check, however I'm fairly certain that in the corner case where we convert numbers exactly between two integers it will round differently, as it does in other cases already.

As I've seen:
* Intel SIMD variants `rounds towards zero` (via truncation)
* MSVC x86 assembly variant depends on FPU rounding mode
* Fallback manual rounding method `rounds towards +∞`
* Most other variants of `float2int` uses `round to nearest, ties to even`.

Using `vcvtaq_s32_f32` intrinsic on AArch64 systems will `round to nearest, ties away from zero`

One solution could be to further extend unit tests already added to check the correctness of the conversions.
Alternatively, we could aim to achieve the more ubiquitous behaviour with a bit of performance penalty.

As far as I understand, however, in the case of digital signal processing, the benefit of the performance uplift of this solution outweighs the occasional mismatch by one on the output and could be acceptable.

Please advise me how to proceed / what would be an acceptable solution for you.

agosdahu:
The new url looks promising, `run_vectors.sh` can finish successfully! Thanks!
I will run the tests in the upcoming days.

celt/arm/celt_neon_intr.c
jmvalin: Yeah, I see the problem. Indeed we don't really care how ties get rounded. Maybe a simple way to do the check would just be to verify that the integer value differs from the input float by less than 0.501 or so?

jmvalin:
So I just landed some changes that appear to break your patch, but it shouldn't be too hard to update. In your patch, you had:
```
-      for (i=0;i<ret*st->channels;i++)
-         pcm[i] = FLOAT2INT16(out[i]);
+      celt_float2int16(out, pcm, ret*st->channels, st->arch);
```

and it now turns out that the code reads:
```
          for (i=0;i<ret*st->channels;i++)
             pcm[i] = RES2INT16(out[i]);
```

The same function is now used for both fixed-point and float, which means that I *believe* your patch should update the code to read something like:
```
#ifdef FIXED_POINT
          for (i=0;i<ret*st->channels;i++)
             pcm[i] = RES2INT16(out[i]);
#else
          celt_float2int16(out, pcm, ret*st->channels, st->arch);
#endif
```

Among the changes that just landed are a new 24-bit integer API, including a opus_decode24() call that decodes to int32 instead of int16. So if you have any bandwidth, it may be interesting to have a separate celt_float2int32() patch to optimize that.

celt/arm/celt_neon_intr.c
agosdahu: I pushed a modification containing a  check to see if the intrinsic implementation is off by maximum 1

agosdahu:
Rebased on top of your modifications and applied the suggested change

jmvalin:
Were you able to run the opus_build_test.sh script with your updated patch?

agosdahu:
> Were you able to run the opus_build_test.sh script with your updated patch?

I was able to, but unfortunately both the patched and main branches produced failing tests.

I used a 32-core Ampere eMAG workstation to run the tests via the “opus_build_tests.sh” script with the changed line as presented below in order to get outputs from as many builds as possible:

```
-seq -w "$nb_tests" | parallel --halt now,fail=10 -j +2 -q ../random_config.sh "build_tests/run_{}" "$configure_dir" "$oldvectors" "$newvectors"
+seq -w "$nb_tests" | parallel -j "$(nproc)" -q ../random_config.sh "build_tests/run_{}" "$configure_dir" "$oldvectors" "$newvectors"
```

I executed 1000-1000 runs each for both the top of the main branch (at commit `c79a9bd1`, which I called a “vanilla” run) and the patched branch.
I deemed a run successful without any problems, when the generated random_config.txt ended in “all tests PASS” and failed if it ended with “check FAIL”.
I also randomly re-ran a few of the runs to ascertain whether the outcome remained unchanged (which it did).

441/1000 vanilla runs passed without any problems
454/1000 patched runs passed without any problems

I did not run a thorough analysis on all the outputs in search for all the causes for failure, but there didn't seem to be a single root cause or at least it wasn't obvious to me.
I checked, however, for any `OPUS_CHECK_ASM` or `celt_assert` related errors and found none.

Please advise me if there is anything more I can do to advance the fate of this PR.

I have a collection of the build outputs, which I initially used to analyse.
Do you want me to share it with you?

I have collected the following four files for each of the runs on both branches and prefixed them with their run ID. (e.g.: `0001_<filename.txt>`):
- `configure_output.txt`
- `make_output.txt`
- `makecheck_output.txt`
- `random_config.txt`
 
I’ve also collected the cflags, the unique config parameters and the testvector sets used for each failing runs and listed them, as well as their run ID, in three different files respectively.

I still have the full build outputs, if you'd need some other logs, I'd be happy to get those as well for you.

jmvalin:
Can you upload the data for the failures you get? I'd need the random_config.txt file and the relevant _output.txt for where the error occurs.

agosdahu:
I'm attaching all failing results for both runs.
I hope it'll be helpful.
[failed_outputs.tar.gz](https://github.com/user-attachments/files/19231016/failed_outputs.tar.gz)

jmvalin:
OK, so I can see mostly two issues in these logs. One seems to be a bug I recently fixed in main (commit 030074f9) that was affecting clang but not gcc/x86. The other is that it appears that -ffast-math is breaking a lot of things, which is strange because I haven't observed that before. So could you try running the tests again on the latest main while removing the use of -ffast-math from random_config.sh ?
Also, just curious what compiler/version are you using?

agosdahu:
Sure, I'll fire up a run on the latest main and with removed -ffast-math and will get back to you with the results.
I used Clang 18.1.3

jmvalin:
Actually, I ran the test scripts with clang-18 and it's indeed finding a bunch of errors (not all of which related to the issues I mentioned above). So let me fix those first. In the mean time, are you able to test with gcc?

jmvalin:
So I *think* I was able to fix all the failures with clang. The last one appears to actually be a compiler bug in the VLA handling (checked in a workaround). So you might want to test that out too.

agosdahu:
It took me about 4 days to finish the 1000 runs with gcc, which was really unexpected.
I ran the tests on the patched branch, which is currently rebased on `2169f64`. 
(I planned to have main tested as well, but it seems I was optimistic about the time required to run them, sorry about that)

Since I forgot to remove the `-ffast-math` flag from my gcc run, I had consistent unit test failures with configuration containing the fast-math flag, outputing error message `/celt/entenc.c, line 209: assertion failed: _bits>0` originating from `tests/test_opus_custom`.

I started a new run with current HEAD of main using clang and the fast-math flag to test with your fixes
If all is well, I'll rebase and retry with the patch as well

jmvalin:
Normally, the testing on the latest main should work even if -ffast-math enabled. It's possible that the assert was caused by the issue that was fixed in 738c29fc, but if you ever see that assertion in celt/entenc.c, line 209, with the recent fixes (with or without -ffast-math), then there's definitely something wrong. 

Regarding the clang bug I mentioned, you can see it at https://github.com/llvm/llvm-project/issues/131671 but the latest main should now work around it.

agosdahu:
Testing on (the then) top of main (`4a344aff`) resulted no errors.
Running tests on rebased patch branch.

agosdahu:
I'm happy to report that the full test suite at `85222a5` has finished with `No error found` after 1000 runs.

jmvalin:
Thanks for running/debugging all these tests. Just landed your PR in main. Any thoughts/interest in a 32-bit version as explained above?

agosdahu:
I'm happy to see the patch landing in main, thanks for all the help.
We have another Arm optimisation patch coming up soon, but that targets the pcm soft clipping function.
Seeing the roadmap we have at the moment, you can expect the implementation of the 32-bit version, but not immediately.

PR #380, opened 2024-12-08
fix test code for detecting avx2 support

proyvind:
This commit fixes AVX2 compiler support detection code in configure.ac.


sezero:
Also see #362 which is intended to fix #361. It is a much smaller diff by comparison to this one.

jmvalin:
Also has the same issue with __m128i_u which isn't portable.

PR #389, opened 2025-02-04
Using autogen.sh allow fallback to curl if wget is missing for build

heitbaum:
Update the build scripts to optionally allow curl instead of wget.


jmvalin:
merged

Issue #368, opened 2024-10-04
RISC-V port implementation


k-kisielak:
I want to announce that we are currently working on RISC-V implementation using vector extension.
As we I mean: Samsung R&D Poland + partially RISE Project engineers

Code is being developed on feature branch at: https://github.com/k-kisielak/opus/tree/rvv_impl
Implementation is in early stage and is not deemed for merging in current form.

**Current state:** 
We started with implementing parts of silk module:
- basing on existing SSE implementation we prepared one using RISC-V vector intrinsics

**Items under development:**
- tests are to be performed, both functional and performance ones - to test those changes we have access to RISC-V devices like BananaPI F3, we have experience with testing RVV under qemu
- we are looking into expanding github action configs to allow testing RISC-V version (as of now only on linux)
- we are integrating more RVV implementations, namely for burg modified algorithm

We would gladly accept any feedback, thank you!

petterreinholdtsen:
[Krystian Kisielak]
> We would gladly accept any feedback, thank you!

I can not really help with the code, but can tell you I am very happy to
read that you are working on an optimized RISC-V implementation.  Keep
up the good work. :)

-- 
Happy hacking
Petter Reinholdtsen

jmvalin:
This is indeed good news. Here's a few questions/suggestions on how to proceed:
1) Are you targeting fixed-point or floating point or both? On some devices one is faster than the other, but it's not always the same. All things being equal float is generally preferable because it has more features.
2) When optimizing, it's best to start with just defining optimized version of some of the basic operations. For example, you can look at celt/arm/fixed_armv4.h and silk/arm/macros_armv5e.h. That's an easy way to optimize many functions at once without changing too much code.
3) If optimizing entire function, replicating something similar to SSE/ARM like you're doing is a good idea. Make sure you implement the CHECK_ASM scheme for debugging with --enable-check-asm
4) Splitting into small patches that each add functionality without regressions (e.g. optimize one function and use it) is the best way to land these changes.

k-kisielak:
Ad.1 We are opting for 'somewhat capable' devices with vector units operating on floats. So fixed point is less important.
Ad. 2-4 Thank you for tour advice, as of now we focused on bulky functions yet we will look for ones which could provide speedup with small changes exploiting vector extension capabilities.

jmvalin:
[...]

PR #392, opened 2025-02-20
RFC: Streamline implementation overrides

MarekPikula:
# Background and motivation

While preparing initial patches for RISC-V support (as mentioned in #368), I noticed that overriding function implementations requires a lot of manual copy-pasting, leading to code duplication and increased maintenance effort.

This PR proposes a streamlined approach to declaring function overrides. I applied it to two functions in Silk (`silk_NSQ` and `silk_VQ_WMat_EC`) for x86 to demonstrate its benefits and gather feedback on whether such a refactor would be accepted upstream.

If approved, I can extend this approach to all overridable functions across all architectures, including Celt (which seems to use a similar override mechanism – please correct me if I'm wrong).

# Proposal

1. To reduce code duplication, function declaration can be set in a single macro (in`main_overrides.h`), which can also be used for function type definition for use in the implementation array (e.g., for `silk_NSQ` function, there is now `SILK_NSQ_DECL(impl)` macro, which can be used to declare the function for all backends).
2. The `arch` argument in the main macro definition is moved to the front, allowing for the use of macro variadic arguments, reducing the copy-paste of argument lists for each implementation.
3. To declare a single backend override, there is a new macro `OVERRIDE_IMPL_SINGLE`, which references the selected backend implementation.
4. Accordingly, for the multi-backend override, there is a new macro `OVERRIDE_IMPL_ARRAY`, which addresses the override array created with `OVERRIDE_IMPL_ARRAY_DECL`.

This approach significantly reduces manual (and error-prone) copy-paste work, minimizes the risk of errors when developing new backends (e.g., RISC-V in our case), and improves maintainability. Now, function definition changes require modification in only one place, reducing the likelihood of regressions across platforms.

## Example

To enable the override of a function `foo` in Silk:
1. Add function declaration in `main_overrides.h` in the form of `FOO_DECL(impl, ...)` macro and function type definition with `typedef FOO_DECL(t, *const);`.
2. Add the declaration to `main.h`:
   ```c
   FOO_DECL(c);
   #ifndef foo
   #define foo(...) OVERRIDE_IMPL_SINGLE(foo, c, __VA_ARGS__)
   #endif
   ```
3. Change the existing function definition to `FOO_DECL(c) { ... }`.
4. Change all function references to include `arch` as the first argument.

Now, you can add arch-specific implementation (an example for x86 SSE):
1. Add function override in `main_sse.h`:
   1. For single function override:
      ```c
      #define foo(...) OVERRIDE_IMPL_SINGLE(foo, sse4_1, __VA_ARGS__)
      ```
   2. For RTCD override:
      ```c
      extern OVERRIDE_IMPL_ARRAY_DECL(foo);
      #define foo(...) OVERRIDE_IMPL_ARRAY(foo, __VA_ARGS__)
      ```
2. Add RTCD map definition in `x86_silk_map.c`:
   ```c
   OVERRIDE_IMPL_ARRAY_DECL(foo) = {
     foo_c,                  /* non-sse */
     foo_c,
     foo_c,
     MAY_HAVE_SSE4_1( foo ), /* sse4.1 */
     MAY_HAVE_SSE4_1( foo )  /* avx */
   };
   ```
3. Add arch-specific function definition in `foo.c` with
   ```c
   FOO_DECL(sse4_1)
   {
       ...
   }
   ```

### DIFF
--- a/silk/NSQ.c
+++ b/silk/NSQ.c
@@ -73,24 +73,7 @@ static OPUS_INLINE void silk_noise_shape_quantizer(
 );
 #endif
 
-void silk_NSQ_c
-(
-    const silk_encoder_state    *psEncC,                                      /* I    Encoder State                   */
-    silk_nsq_state              *NSQ,                                         /* I/O  NSQ state                       */
-    SideInfoIndices             *psIndices,                                   /* I/O  Quantization Indices            */
-    const opus_int16            x16[],                                        /* I    Input                           */
-    opus_int8                   pulses[],                                     /* O    Quantized pulse signal          */
-    const opus_int16            *PredCoef_Q12,                                /* I    Short term prediction coefs     */
-    const opus_int16            LTPCoef_Q14[ LTP_ORDER * MAX_NB_SUBFR ],      /* I    Long term prediction coefs      */
-    const opus_int16            AR_Q13[ MAX_NB_SUBFR * MAX_SHAPE_LPC_ORDER ], /* I    Noise shaping coefs             */
-    const opus_int              HarmShapeGain_Q14[ MAX_NB_SUBFR ],            /* I    Long term shaping coefs         */
-    const opus_int              Tilt_Q14[ MAX_NB_SUBFR ],                     /* I    Spectral tilt                   */
-    const opus_int32            LF_shp_Q14[ MAX_NB_SUBFR ],                   /* I    Low frequency shaping coefs     */
-    const opus_int32            Gains_Q16[ MAX_NB_SUBFR ],                    /* I    Quantization step sizes         */
-    const opus_int              pitchL[ MAX_NB_SUBFR ],                       /* I    Pitch lags                      */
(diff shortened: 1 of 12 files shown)


MarekPikula:
Hi @jmvalin @rillian, could you take a look at this (or ping some other person who could make a comment)?

jmvalin:
I didn't actually create that function pointer setup, so I'm not sure of all the reasoning behind it. What I'd suggest though is to make sure you run the randomized build test to check that your patch doesn't break anything. See my previous comment at https://github.com/xiph/opus/pull/379#issuecomment-2637381867 for more details.

MarekPikula:
Thanks for a comment @mark4o. I didn't realize C90 was the required standard for Opus. Maybe it would be nice to enforce it with `--std=c90` in the build config?
I have an idea how to do the refactor without `__VA__ARGS__`. I'll push a new proposal later this week.

PR #396, opened 2025-03-28
Arm: Speed up -1..1 soft clipping with Neon

agosdahu:
If the signal exceeds -1..1 then, as error handling, the soft_clip function forces the signal back into -1..1. This is problematic since the search loop to find the next sample exceeding -1..1 is slow. If cheap on the current platform, while doing -2..2 hardclipping we can also detect if the signal never exceeds -1..1, avoiding the need for a second search loop.

### DIFF
--- a/celt/arm/celt_neon_intr.c
+++ b/celt/arm/celt_neon_intr.c
@@ -86,7 +86,84 @@ void celt_float2int16_neon(const float * OPUS_RESTRICT in, short * OPUS_RESTRICT
       out[i] = FLOAT2INT16(in[i]);
    }
 }
+
+int opus_limit2_checkwithin1_neon(float *samples, int cnt)
+{
+   const float hardclipMin = -2.0f;
+   const float hardclipMax = 2.0f;
+
+   int i = 0;
+   int exceeding1 = 0;
+   int nextIndex = 0;
+
+#if defined(__ARM_NEON)
+   const int BLOCK_SIZE = 16;
+   const int blockedSize = cnt / BLOCK_SIZE * BLOCK_SIZE;
+
+   float32x4_t min_all_0 = vdupq_n_f32(0.0f);
+   float32x4_t min_all_1 = vdupq_n_f32(0.0f);
+   float32x4_t max_all_0 = vdupq_n_f32(0.0f);
+   float32x4_t max_all_1 = vdupq_n_f32(0.0f);
+
+   float max, min;
+
+   for (i = 0; i < blockedSize; i += BLOCK_SIZE)
+   {
+      const float32x4_t orig_a = vld1q_f32(&samples[i +  0]);
+      const float32x4_t orig_b = vld1q_f32(&samples[i +  4]);
+      const float32x4_t orig_c = vld1q_f32(&samples[i +  8]);
+      const float32x4_t orig_d = vld1q_f32(&samples[i + 12]);
+      max_all_0 = vmaxq_f32(max_all_0, vmaxq_f32(orig_a, orig_b));
+      max_all_1 = vmaxq_f32(max_all_1, vmaxq_f32(orig_c, orig_d));
+      min_all_0 = vminq_f32(min_all_0, vminq_f32(orig_a, orig_b));
+      min_all_1 = vminq_f32(min_all_1, vminq_f32(orig_c, orig_d));
+   }
+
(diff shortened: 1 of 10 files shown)


agosdahu:
This patch was created concurrently with the float-to-int16 optimisation patch and **all uplift result was measured on top of it**.

When measuring performance uplift, the same methods were used:
- method A) measuring mono/stereo output at 48k sampling via `run_vectors.sh` on test vector set attached to repo,
- method B) running 100-100 stereo run at 48k sampling rate for test files testvector09.bit and testvector10.bit, respectively, but in succession.

The test target was a single Cortex A55 core of a Google Tensor G2 SoC.

Results:
| **method A**        | clang 18.1.8 | gcc 14.2 |
|---------------------|--------------|----------|
| using /usr/bin/time | 3%           | 6%       |
| using linux perf    | 5%           | 8%       |

| **method B**        | clang 18.1.8 | gcc 14.2 |
|---------------------|--------------|----------|
| using /usr/bin/time | 4%           | 6%       |
| using linux perf    | 4%           | 6%       |

I also ran the tests described in previously in [PR#379](https://github.com/xiph/opus/pull/379#issuecomment-2654060631) and found no error.

src/opus.c
```diff
@@ -30,23 +31,30 @@
 #endif
 
 #include "opus.h"
+#include "celt/mathops.h"
 #include "opus_private.h"
 
 #ifndef DISABLE_FLOAT_API
-OPUS_EXPORT void opus_pcm_soft_clip(float *_x, int N, int C, float *declip_mem)
+
+static void opus_pcm_soft_clip_impl(float *_x, int N, int C, float *declip_mem, int use_arch, int arch)
```
jmvalin, line 39: Why do you need both use_arch and arch, instead of just setting arch=0 to get the C version?

celt/arm/celt_neon_intr.c
```diff
@@ -86,8 +86,90 @@ void celt_float2int16_neon(const float * OPUS_RESTRICT in, short * OPUS_RESTRICT
...
+   const float hardclipMin = -2.0f;
+   const float hardclipMax = 2.0f;
+
+   int i = 0;
+   int exceeding1 = 0;
+   int start = 0;
+
+#if defined(__ARM_NEON)
+   const int BLOCK_SIZE = 16;
+   const int FRAME_SIZE = (60000 / sizeof(float) / BLOCK_SIZE) * BLOCK_SIZE;
```
jmvalin, line 101: I'm not sure I get this FRAME_SIZE. Looks like a cache optimization thing? In any case, given that Opus packets cannot be longer than 120 ms, I don't think that FRAME_SIZE value can ever be reached unless opus_pcm_soft_clip() gets called explicitly. Or did I miss something here?

src/opus.c
agosdahu: I removed use_arch and adjusted the relevant parts.
Arch is always passed and when it's 0, it is going to yield the C implementation.

celt/arm/celt_neon_intr.c
agosdahu: Originally, yes, it aimed to keep sample processing within cache limits, but we missed the maximum of the opus packet size.
I removed `FRAME_SIZE` and relevant logic handling it.

src/opus.c
```diff
@@ -135,6 +147,17 @@ OPUS_EXPORT void opus_pcm_soft_clip(float *_x, int N, int C, float *declip_mem)
       declip_mem[c] = a;
    }
 }
+
+void opus_pcm_soft_clip_with_arch(float *_x, int N, int C, float *declip_mem, int arch)
```
jmvalin, line 151: This doesn't seem very useful. You might as well just call the impl() function from everywhere, instead of the _with_arch() one that then calls impl().

jmvalin:
Posted one more comment. Would be good to add some comments about all_within_neg1pos1 and opus_limit2_checkwithin1() to explicitly state what the purpose of the function/variable is. Especially since if you don't look at the ARM version, the code looks a bit silly.

agosdahu:
I added a lengthy explanation (I hope you had something like this in mind) and removed the *_with_arch() calls, using only the *_impl() ones.

jmvalin:
merged

agosdahu:
Thank you Jean-Marc!
Much appreciated.

PR #398, opened 2025-04-18
add tar -z option

jg1uaa:
only GNU tar recognizes compressed file automatically.
please do not forget "-z" option with tar when handling .tar.gz file, for portability.


rillian (approved):
Seems reasonable to me. Just means we have to be careful if we switch to another compression method later.

jmvalin:
merged

PR #449, opened 2026-01-10
MSVC ARM64/ARM64EC support

DEATH923:
Fixes compilation errors for MS ARM targets (cannot use xmmintrin.h on ARM64EC).

Now builds cleanly for ARM64 / ARM64EC via CMake, automatically enables Neon optimizations for Windows ARM targets.

These changes are already in use in foobar2000 so you can consider them well tested - non MSVC configs should not be affected by this.


xnorpx:
@jmvalin changes looks good to me, @DEATH923 says it is well tested.

Version 1.5/1.6 is WOA Intrinsics is broken so this is a good change.

PR #457, opened 2026-01-30
cmake: define ENABLE_OSCE_BWE when OPUS_OSCE is set

PiMaker:
This fixes compilation of the bandwidth extension module when compiling with cmake.

Matches the behaviour of `--enable-osce` from https://github.com/xiph/opus/blob/main/configure.ac#L966 which also ties the 2 options together.


xnorpx:
@jmvalin change looks good if we want to bundle these things together.

PR #464, opened 2026-03-20
Init 'up' to fix clang-cl uninitialized warning

jim-signal:
Building for Windows in the WebRTC context uses clang-cl (Clang's MSVC-compatible driver), which doesn't recognize the GCC-style noreturn attribute. The default branch is treated as potentially returning, and warns that `up` may be uninitialized.


jmvalin (approved):

jim-signal:
Merged

PR #465, opened 2026-03-22
osce: Clamp pitch embedding index

matejsmycka:
Missing bounds clamp on SILK pitch lag before OSCE embedding table lookup causes up to 7936-byte heap OOB read when the model is trained with Python default pitch_max=257 (max lag is 288); fix adds IMIN/IMAX clamping matching FARGAN's existing pattern.


janpbuethe (approved):
The included models cover the entire SILK period range but it makes sense to have this safety mechanism.

PR #473, opened 2026-05-06
Optimize Silk encoder NEON paths and fix warped autocorrelation precision bug

wangzihao3:
## Summary

  This PR applies targeted ARM NEON optimizations to two Silk encoder hot paths (~25% speedup, ~30% fewer instructions), and fixes a precision loss bug in the warped
   autocorrelation computation.

  ### Changes

  **silk/LPC_analysis_filter.c — New NEON implementation**
  - Add a NEON-vectorized `silk_LPC_analysis_filter()` that processes 8 filter taps per iteration using paired `vmlal_s16` accumulators (`int32x4_t`).
  - The original scalar implementation is preserved as `silk_LPC_analysis_filter_c()` for reference and `OPUS_CHECK_ASM` validation.

  **silk/arm/NSQ_del_dec_neon_intr.c — Multiple optimizations**
  - **LPC prediction**: Replace 16 individual `vld1q_s32` loads with `ldp` (load-pair) inline assembly, loading 8 data vectors + 2 coefficient vectors per block. Use
   `vqdmulhq_laneq_s32` to index directly into 128-bit coefficient vectors, eliminating `vget_low/high_s32` split overhead.
  - **Allpass sections**: Fully unroll the allpass loop via macro with compile-time branching on `shapingLPCOrder`, enabling precise register allocation and removing
   loop overhead.
  - **Common subexpression elimination**: Precompute `Tilt_Q14_Q16` and `LF_shp_Q14_Q15` outside the per-sample loop; cache `LF_AR_Q14` to avoid redundant vector
  loads.
  - **AR_shp_Q28 widening**: Remove the scalar tail loop since `MAX_SHAPE_LPC_ORDER` (24) is exactly divisible by 8.
  - **State replacement copy**: Unroll the `numOthers` copy loop 16× with `__builtin_prefetch` to hide memory latency during the candidate state swap.

  **silk/fixed/arm/warped_autocorrelation_FIX_neon_intr.c — Fix precision loss in autocorrelation**
  - The original code computed `corr_QC[orderT]` (the zero-lag autocorrelation) from the raw `int16` input using `vmull_s16`, then left-shifted the result by `QC`.
  This lost precision because it operated on the un-warped source data instead of the warped signal that the rest of the function processes.
  - The fix computes `corr_QC[orderT]` from the already-warped `input_QS` (int32) data using `vmull_s32` + `vsraq_n_s64` with the correct right shift `2*QS - QC`,
  consistent with how all other `corr_QC[]` entries are accumulated. This ensures the autocorrelation is derived from the same warped signal path, eliminating the
  precision discrepancy.

  ### Performance

  | Metric | Improvement |
  |--------|------------|
  | Encoding speed | ~25% faster |
  | Dynamic instruction count | ~30% fewer |

### DIFF
--- a/silk/LPC_analysis_filter.c
+++ b/silk/LPC_analysis_filter.c
@@ -31,6 +31,8 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #include "SigProc_FIX.h"
 #include "celt_lpc.h"
+#include <arm_neon.h>
+#include "stack_alloc.h"
 
 /*******************************************/
 /* LPC analysis filter                     */
@@ -46,7 +48,7 @@ POSSIBILITY OF SUCH DAMAGE.
    C89-compliant. */
 #define USE_CELT_FIR 0
 
-void silk_LPC_analysis_filter(
+void silk_LPC_analysis_filter_c(
     opus_int16                  *out,               /* O    Output signal                                               */
     const opus_int16            *in,                /* I    Input signal                                                */
     const opus_int16            *B,                 /* I    MA prediction coefficients, Q12 [order]                     */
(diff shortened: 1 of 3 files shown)


wangzihao3:
Hi, @jmvalin @rillian. Could someone please help review this PR when you have a moment? It includes targeted ARM NEON optimizations for the Silk encoder and fixes a precision loss bug in the warped autocorrelation computation.
Please let me know if you need any additional information or if there are any changes required. Thanks in advance for your time!

rl123567:
Hi, @jmvalin @rillian. Could someone please help review this PR when you have a moment? It includes targeted ARM NEON optimizations for the Silk encoder and fixes a precision loss bug in the warped autocorrelation computation.
Please let me know if you need any additional information or if there are any changes required. Thanks in advance for your time!

jmvalin:
This commit appears to be doing many things at once and it's not clear what is what. It claims to fix a bug, seems to add a new function, and modifies existing ones in ways that seem a bit dangerous (assuming length is a multiple of 8). I'd first suggest breaking it down into at *least* 3 parts (if not more) with each patch doing one isolated thing that improves things without breaking anything else.

PR #471, opened 2026-04-28
Gate ARM DNN RTCD source on DNN builds

cookesan:
Summary:
- Only add the ARM DNN runtime-dispatch source when DNN sources are enabled.
- Match the x86 CMake path and Autotools gating so non-DNN ARM builds do not compile arm_dnn_map.c.

Testing:
- git diff --check
- cmake -S . -B build-local -DOPUS_BUILD_TESTING=ON -DOPUS_BUILD_PROGRAMS=ON
- cmake --build build-local --parallel
- ctest --test-dir build-local --output-on-failure

Fixes #443.

PR #482, opened 2026-06-16
cmake: build the Arm DOTPROD DNN kernels (vdotq_s32)

czoli1976:
## Summary

CMake builds never compiled the int8 dot-product DNN tier. `cmake/OpusSources.cmake` reads `DNN_SOURCES_DOTPROD` into `${dnn_sources_arm_dotprod}`, but that variable was never added to a target, and CMake had **no Arm DOTPROD detection at all**. So CMake builds with DNN features (DRED / OSCE / Deep PLC) ran the multi-instruction `vmull_s8` / `vpadalq_s16` emulation in `cgemv8x4` instead of the one-instruction `vdotq_s32` — unlike the Autotools and Meson builds, which wire it correctly.

## Fix

- `opus_detect_dotprod()` in `OpusFunctions.cmake` — a compiler test for `vdotq_s32` under `-march=armv8.2-a+dotprod`, mirroring `opus_detect_neon()`.
- `OpusConfig.cmake` — call it on Arm and set `OPUS_MAY_HAVE_DOTPROD` when supported.
- `CMakeLists.txt` — when supported, define `OPUS_ARM_MAY_HAVE_DOTPROD` and (for DNN builds) compile `${dnn_sources_arm_dotprod}` with the dotprod `-march` flag.

The kernel is dispatched at runtime via the existing RTCD table; non-DOTPROD targets keep the NEON path. This brings CMake in line with Autotools/Meson.

## Verification (Apple M4)
- `cmake -DOPUS_DRED=ON`: configure reports `Check DOTPROD support by compiler ... Success`, `nnet_dotprod.c` compiles with `-march=armv8.2-a+dotprod`, and `compute_linear_dotprod` is linked (previously absent).
- Default (non-DNN) build unaffected.

Independent of, and complementary to, #471 / #478 (which gate the Arm DNN RTCD source / fix a Windows-ARM linker issue when DNN is *disabled*); this adds the DOTPROD tier when DNN is *enabled*.


jmvalin:
Anyone can review this? @xnorpx ?

PR #488, opened 2026-07-14
silk/arm/NSQ_del_dec_neon_intr.c — Multiple optimizations

rl123567:
**silk/arm/NSQ_del_dec_neon_intr.c — Multiple optimizations**

- LPC prediction: Replace 16 individual vld1q_s32 loads with ldp (load-pair) inline assembly, loading 8 data vectors + 2 coefficient vectors per block. Use vqdmulhq_laneq_s32 to index directly into 128-bit coefficient vectors, eliminating vget_low/high_s32 split overhead.
- Allpass sections: Fully unroll the allpass loop via macro with compile-time branching on shapingLPCOrder, enabling precise register allocation and removing loop overhead.
- Common subexpression elimination: Precompute Tilt_Q14_Q16 and LF_shp_Q14_Q15 outside the per-sample loop; cache LF_AR_Q14 to avoid redundant vector loads.
- AR_shp_Q28 widening: Remove the scalar tail loop since MAX_SHAPE_LPC_ORDER (24) is exactly divisible by 8.
- State replacement copy: Unroll the numOthers copy loop 16× with __builtin_prefetch to hide memory latency during the candidate state swap.

### DIFF
--- a/silk/arm/NSQ_del_dec_neon_intr.c
+++ b/silk/arm/NSQ_del_dec_neon_intr.c
@@ -508,35 +508,93 @@ static OPUS_INLINE int32x4_t silk_SMLAWB_lane1_neon(
     return vaddq_s32( out_s32x4, vqdmulhq_lane_s32( in_s32x4, coef_s32x2, 1 ) );
 }
 
+static OPUS_INLINE int32x4_t silk_SMLAWB_lane_0_neon(
+    const int32x4_t out_s32x4,
+    const int32x4_t in_s32x4,
+    const int32x4_t coef_s32x4
+)
+{
+    return vaddq_s32( out_s32x4, vqdmulhq_laneq_s32( in_s32x4, coef_s32x4, 0 ) );
+}
+
+static OPUS_INLINE int32x4_t silk_SMLAWB_lane_1_neon(
+    const int32x4_t out_s32x4,
+    const int32x4_t in_s32x4,
+    const int32x4_t coef_s32x4
+)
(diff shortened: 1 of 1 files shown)


rl123567:
Hi, @jmvalin Could you please help review this PR when you have a moment? It includes targeted NSQ_del_dec_neon_intr.c Multiple optimizations，for details, see the commit message. Please let me know if you need any additional information or if there are any changes required. Thanks in advance for your time!

lpi:
Tested head: `330eae9f51f144ec195000e5554d0ebbef8724f9`.

I tested safe ideas independently on top of #503. Their approximate combined production-c10 encoder improvements versus stock were:

- paired/laneq: **+4.0820% CPU / +4.0825% wall**
- Tilt/LF hoists: **+4.0543% / +4.0692%**
- LF_AR cache: **+4.1450% / +4.1417%**
- #503 alone: **+4.0713% / +4.0826%**

Positive means faster/lower per-call cost. These stock-relative numbers are compounded cross-run estimates, not direct stock arms. Only LF_AR caching added a small increment over #503: **+0.0768% CPU / +0.0616% wall**.

I did not time the submitted bundle wholesale because its inline assembly does not model pointed-to memory, its prefetch pointer formation extends beyond the derived state-row region, and its state-copy rewrite overlaps #503. The allpass-unroll and AR-tail ideas compiled to equivalent Release code. The tested isolated variants were output-exact and sanitizer-clean.

PR #492, opened 2026-08-17
projection: add ambisonic mapping matrices for 6th to 14th order

abhayap:
Channel mapping family 3 already accepts up to 227 channels, but the built-in projection matrices stop at 5th order, so anything above that fails in opus_projection_ambisonics_encoder_create(). Add the missing tables for 6th to 14th order in a new src/mapping_matrix_hoa.c.

These are not built the way the 4th- and 5th-order tables are. Those sample the spherical-harmonic matrix at Fliege-Maier cubature nodes and invert it in the least-squares sense, which is fine to 9th order and then degrades badly: cond(A) goes 3.8 at 9th order, 86 at 10th, 723 at 12th, because those node sets are chosen to be good cubature rules rather than good interpolation points, and the 121-point set is the first with a negative weight. Conditioning is what the demixing step costs in SNR -- per-sector coding noise returns ||A||_F.||A^-1||_F/N times louder in the ambisonic domain -- so extending that construction would give up 15 dB at 10th order and 32 dB at 12th.

Instead these sample at maximum-determinant (Fekete) nodes, which come in sets of exactly (N+1)^2 points chosen to maximize |det Y|, and take the polar factor of the result: the nearest matrix with orthonormal columns. That is energy-preserving ambisonic decoding (Zotter, Pomberger & Noisternig, Acta Acustica 98(1), 2012) at the critically sampled limit L = (N+1)^2. At every order it gives cond(A) = 1, needs no gain field, makes the demixing matrix the exact transpose of the mixing matrix, and holds the Q15 round-trip error at the quantization floor instead of letting it grow with order.

Also:

  - Replace the per-order if/else chains in opus_projection_encoder.c with mapping_matrix_get_ambisonic(), so adding an order is a table entry rather than four more branches in two places.

  - Drop the 65004-octet cap in mapping_matrix_get_size(). That is the most a mapping matrix can be and still fit an Ogg OpusHead, which is a container limit for the muxer to enforce, not an allocator limit; it excludes only 13th and 14th order (78 KB and 103 KB of cells). The 255-channel bound stays.

  - Extend tests/test_opus_projection.c: check that each built-in pair inverts itself to within Q15 (46-64 dB, on a diffuse field), and that a synthetic ambisonic signal -- four tonal sources encoded as plane waves in ACN/SN3D -- survives encode and decode at every order, with and without the non-diegetic channels (~27 dB at 64 kbit/s per channel). The codec figure averages two sets of source directions, because how sparsely four sources happen to land on the sectors moves it by 1-2 dB either way, which is more than the matrices differ by. The synthesis is deterministic and self-contained; no test vectors are needed.


jmvalin:
I had a look and so far everything looks sane to me. The only real issue is the tables themselves, which would add about 700 kB to the Opus library, pretty much doubling its size for a feature that few people will use. At this point, I think it would be best to have a build-time option that specifies the max ambisonics order to support in the encoder (IIRC the decoder does not need the tables).

abhayap:
Superseded by #499, which answers the review here in two ways.

The tables are halved rather than just capped: all 14 orders are rebuilt with one construction that makes the mixing matrix orthogonal, so the demixing matrix is its exact transpose and is derived at init time instead of stored. That is 358 KiB for all 14 orders against the ~716 KiB two-tables-per-order version you were looking at.

On top of that, `OPUS_AMBISONICS_MAX_ORDER` (1 to 14, default 5) gates which tables get compiled in, wired through autotools, CMake and meson. Your recollection about the decoder was right and is now stated in the header: it takes the demixing matrix from the OpusHead per RFC 8486 §3.2 and never reads these tables, so decoding is unaffected by the setting. The default build actually comes out 6.1 KiB *smaller* than today, since today ships both a mixing and a demixing table for each of orders 1 to 5.

Also worth noting for anyone reading back: the 9 red Windows/CMake jobs on this PR were not from the change. The `CMake` workflow was failing on plain `main` and on unrelated PRs in the same window, until "Update Windows CI for Visual Studio 2026" (a6128f40) landed.

PR #493, opened 2026-08-23
silk: fix rounding-shift overflow in NSQ_del_dec AVX2

fippo:
silk_sar_round_32() and silk_mm_srai_round_epi32() round by adding 1 << (bits-1) before shifting. Their callers pass values straight out of silk_SUB32_ovflw() and the saturating add/sub helpers, so the operand can sit within the bias of INT32_MAX and the addition overflows.

For the scalar helper that is undefined behaviour, reported by UBSAN:

  NSQ_del_dec_avx2.c:103:7: runtime error: signed integer overflow:
  2147481683 + 2048 cannot be represented in type 'int'

reached via silk_LPC_analysis_filter_avx2(), which feeds silk_sar_round_32(out32_Q12, 12) with the result of silk_SUB32_ovflw(). The SIMD helper wraps rather than being UB, but has the same defect.

In both cases the wrap also breaks bit-exactness with the C code, which rounds via silk_RSHIFT_ROUND() -- written as ((a >> (shift-1)) + 1) >> 1 precisely so it cannot overflow. On the reported input the C path yields +524288 (saturating to 32767) while AVX2 yields -524288 (saturating to -32768), i.e. the same input encodes differently depending on whether the CPU has AVX2.

Round the same way the C code does. The two forms are identical whenever the biased add does not overflow, so this changes no other output.

Verified with OPUS_CHECK_ASM + ENABLE_ASSERTIONS: the reported input fails the silk_NSQ_del_dec_c bit-exactness assert before this change and passes after, as does a 6.8M-execution fuzzing run.

Originally found by clusterfuzz, backport of
  https://chromium-review.googlesource.com/c/chromium/src/+/8178247

### DIFF
--- a/silk/x86/NSQ_del_dec_avx2.c
+++ b/silk/x86/NSQ_del_dec_avx2.c
@@ -97,13 +97,6 @@ static OPUS_INLINE opus_int16 silk_sat16(opus_int32 num)
     return num;
 }
 
-static OPUS_INLINE opus_int32 silk_sar_round_32(opus_int32 a, int bits)
-{
-    silk_assert(bits > 0 && bits < 31);
-    a += 1 << (bits-1);
-    return a >> bits;
-}
-
 static OPUS_INLINE opus_int64 silk_sar_round_smulww(opus_int32 a, opus_int32 b, int bits)
 {
 #ifndef OPUS_CHECK_ASM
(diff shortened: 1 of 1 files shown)


fippo:
copying from the chromium CL for papertrail:
```
perf result using the left channel of far48_stereo.pcm from libwebrtc repeated for 21 minutes
and opus_demo silk-only (voip, WB, 24 kb/s, complexity 10):
92,493,020,648 user instructions before the change,
92,518,152,074 after
so 0.027% increase and 400 extra instructions per frame.

jmvalin:
Thanks. PR merged in main.

PR #496, opened 2026-09-06
SSE4.1 and NEON versions of the SILK resampler interpolation

fippo:
`silk_resampler_private_IIR_FIR_INTERPOL()` resamples the SILK output
to the API rate, so it runs for every stream with a SILK layer and
amounts to 15% of a SILK-only and 10% of a hybrid decode.

Each output sample is an 8-tap dot product over 16-bit inputs, which
can be simplified as one 128-bit multiply-add and vectorized.
Rename it to `_c()`, add the usual `OVERRIDE_`/`_IMPL` dispatch and the
two kernels. `silk_resampler_init` gains an architecture argument.

`opus_demo -d`, 48 kHz mono, x86-64 gcc -O3, by TOC mode:

| mode         | interpolation | total |
| ------------ | ------------- | ----- |
| SILK-only WB | 16.3% -> 7.5% | -9.5% |
| hybrid FB    | 10.8% -> 4.8% | -6.3% |
| CELT-only FB | n/a           | n/a   |

Bit-identical change


jmvalin:
Can you rebase on top of main so that all the Windows workers run again?

lpi:
Tested head: `7e65573cc361fd680dc8acbf8dd74c27b524bb5e`.

Route counters showed that standard 8 kHz Opus→PCMU decode did **not** enter the modified interpolation leaf, while the 16 kHz Opus→Opus re-encode decode leg did. The descriptive complete-transcode result versus the qualified Candidate-3/#503 baseline was **+0.1745% CPU / +0.1921% wall** (positive means faster/lower cost); both confidence intervals were positive, but the result was below the 0.75% threshold. The 48 kHz SILK/hybrid paths were diagnostic-only. The production encode row was a no-effect control, so its roughly -0.1% movement should not be attributed to this decoder mechanism. Exact integer-output checks passed.

The earlier valid Candidate-3 result is separate from this later multi-candidate campaign. The later run used AWS Graviton4 `c8gn.large` (Neoverse-V2), GCC 11.5, fixed CPU 1, zero migrations/steal, seven workloads, 28 randomized complete blocks, and 100 warmup/1,000 measured calls per process. Stock sham passed all 392 rows with exact identity, maximum CPU drift 0.0910%, and maximum CPU CI half-width 0.3187%. The 1,372 candidate rows had zero correctness, route, precision, or median-control failures; all target bit-exact rows preserved packet/final-range/decoded-PCM identity. The campaign was nevertheless `GRAVITON4_INVALID` because 11 no-effect decoder controls exceeded a predeclared 2% wall-p99.9 bound. No rerun, filtering, or exclusions were used.

Therefore the numbers are descriptive/non-promotable. FULL was **+0.2915% CPU / +0.3054% wall**, below 0.75% regardless; no promotion decision comes from this run.

PR #499, opened 2026-09-07
projection: ambisonic mapping matrices for 1st to 14th order, behind a build-time order limit

abhayap:
Supersedes #492, which added 6th- to 14th-order projection matrices. @jmvalin's review there was that the tables would add ~700 kB to libopus, roughly doubling it for a feature few people use, and asked for a build-time option capping the maximum ambisonics order in the encoder. This series does both halves of that: it halves the tables outright, and puts what remains behind an order limit that defaults low.

Two commits.

**`projection: rebuild the ambisonic mapping matrices for 1st to 14th order`**

Rather than extending the existing construction to the higher orders, this rebuilds all 14 with one construction: maximum-determinant (Fekete) nodes at the critically sampled limit L = (N+1)², with the sampled spherical-harmonic matrix replaced by its polar factor — energy-preserving ambisonic decoding (Zotter, Pomberger & Noisternig, Acta Acustica 98(1), 2012). Every order gets cond(A) = 1, no gain field, and no SNR given away in the demixing step.

Two consequences that matter here:

- **The demixing matrix no longer needs storing.** An orthogonal mixing matrix inverts to its own transpose, so `mapping_matrix_init_transposed()` derives it. That is what halves the tables: 358 KiB for all 14 orders, against 716 KiB for two tables per order.
- **2nd order gets fixed on the way past.** `mapping_matrix_soa_*` as shipped has cond(A) = 11.7 against 1.3–3.3 either side and gives up ~12 dB in demixing. Through libopus with a binaural render, PEAQ (ITU-R BS.1387) puts that at +2.4 to +2.9 ODG: −3.48 ("annoying") at 64 kbit/s/ch where a well-conditioned matrix reaches −0.56.

No bitstream change, and no compatibility concern: RFC 8486 §3.2 carries the demixing matrix in the OpusHead, so every family-3 file travels with the matrix it was encoded against, and the mixing matrix is encoder-side and never transmitted.

**`projection: make the highest ambisonic order a build-time option`**

`OPUS_AMBISONICS_MAX_ORDER`, 1 to 14, default 5, wired through all three build systems:

```
--enable-ambisonics-max-order=N     # autotools
-DOPUS_AMBISONICS_MAX_ORDER=N       # CMake
-Dambisonics-max-order=N            # meson
```

Measured with clang -O2, static `libopus.a`, no DNN:

| build | size | vs. today |
|---|---|---|
| order 1 | 668.0 KiB | −12.3 KiB |
| **order 5 (default)** | **674.1 KiB** | **−6.1 KiB** |
| order 14 | 1030.1 KiB | +349.9 KiB |

The default build comes out *smaller* than today's, not merely the same: today ships a mixing and a demixing table for each of orders 1–5 (10744 octets), and this stores only the mixing matrix (5372 octets). So a build that doesn't opt in pays nothing for the new orders and gets a small win.

Your recollection about the decoder is right, and it's now stated in the header: the decoder takes the demixing matrix from the OpusHead and never reads these tables, so it decodes any order no matter how the limit is set. `opus_projection_decoder.c` is untouched by either commit. An encoder asked for an order above the limit fails the way an unsupported order already did — `..._get_size()` returns 0, `..._create()` returns `OPUS_BAD_ARG`.

Since the default no longer compiles the high-order tables at all, two CI configurations build all 14: an autotools one that runs `tests/test_opus_projection` (2.8 s at 14th order), and an MSVC one, that being the compiler most likely to object to a 353 KiB translation unit.

The tests cover each built-in pair inverting to within Q15 on a diffuse field (46–65 dB), the derived demixing matrix being the bit-exact transpose of its mixing matrix, and a synthetic ambisonic signal surviving encode and decode at every order with and without the non-diegetic channels (~27 dB at 64 kbit/s/ch). Deterministic and self-contained; no test vectors needed.

One note on #492's CI: the 9 red Windows/CMake jobs there were not from that PR. The `CMake` workflow was failing on plain `main` in the same window (e.g. 8f39f972), and on unrelated PRs, until "Update Windows CI for Visual Studio 2026" (a6128f40) landed. This branch is rebased on top of that.


abhayap:
Since the tables are 23k lines of opaque integers, here is a self-contained script that regenerates every one of them from first principles, so nobody has to take the construction on faith. numpy is the only dependency; it downloads 160 KiB of node data and takes about 20 s cold, 0.15 s warm.

Run it with a path to this branch's `src/` and it checks its output against the committed tables:

```
$ python3 gen_opus_projection_matrices.py path/to/opus/src
order  nodes  cond(A)  SNR loss  min sep  Q15 orthog.  octets  vs. committed
    1      4   1.0000      0.00   109.47        89.7      72  match
    2      9   1.0000      0.00    69.49        87.8     242  match
    3     16   1.0000      0.00    49.61        85.8     648  match
    4     25   1.0000      0.00    39.49        83.5    1458  match
    5     36   1.0000      0.00    32.06        82.4    2888  match
    6     49   1.0000      0.00    27.14        80.8    5202  match
    7     64   1.0000      0.00    24.09        79.9    8712  match
    8     81   1.0000      0.00    21.15        78.9   13778  match
    9    100   1.0000      0.00    19.30        78.1   20808  match
   10    121   1.0000      0.00    17.37        77.4   30258  match
   11    144   1.0000      0.00    16.28        76.5   42632  match
   12    169   1.0000      0.00    14.67        75.7   58482  match
   13    196   1.0000      0.00    13.65        75.1   78408  match
   14    225   1.0000      0.00    12.39        74.6  103058  match
366646 octets of .rodata for all 14 orders (358.1 KiB)
```

All 14 match bit-exactly.

<details>
<summary><code>gen_opus_projection_matrices.py</code></summary>

```python
#!/usr/bin/env python3
"""Regenerate the Opus projection mixing matrices (orders 1-14) from scratch.

Downloads R. Womersley's maximum-determinant (Fekete) node sets, builds the real
orthonormal SH matrix at those nodes, takes its polar factor, and quantizes to Q15.
Prints per-order statistics; with a path to an opus src/ it also checks the result
against the committed tables.

    python3 gen_opus_projection_matrices.py [path/to/opus/src]

Needs numpy. Nodes are cached in ./maxdet-nodes/.
"""
import os, sys, urllib.request, numpy as np

URL = "https://web.maths.unsw.edu.au/~rsw/Sphere/S2Pts/MD/md%03d.%05d"
CACHE = "maxdet-nodes"


def nodes(N):
    """Womersley maxDet node set of degree N: (N+1)^2 unit vectors, in file order."""
    name = "md%03d.%05d" % (N, (N + 1) ** 2)
    path = os.path.join(CACHE, name)
    if not os.path.exists(path):
        os.makedirs(CACHE, exist_ok=True)
        urllib.request.urlretrieve(URL % (N, (N + 1) ** 2), path)
    xyzw = np.loadtxt(path)
    assert xyzw.shape[0] == (N + 1) ** 2, name
    return xyzw[:, :3]


def sh(N, xyz):
    """Real SH at `xyz`, orthonormal, ACN order, no Condon-Shortley: (npts, (N+1)^2)."""
    ct = xyz[:, 2]                                     # cos(inclination)
    st = np.hypot(xyz[:, 0], xyz[:, 1])                # sin(inclination)
    phi = np.arctan2(xyz[:, 1], xyz[:, 0])
    P = {}                                             # fully normalized Legendre
    P[(0, 0)] = np.full(len(ct), 1.0 / np.sqrt(4 * np.pi))
    for m in range(1, N + 1):                          # sectorial
        P[(m, m)] = np.sqrt((2 * m + 1) / (2 * m)) * st * P[(m - 1, m - 1)]
    for m in range(0, N):
        P[(m + 1, m)] = np.sqrt(2 * m + 3) * ct * P[(m, m)]
    for m in range(0, N + 1):
        for n in range(m + 2, N + 1):
            a = np.sqrt((4 * n * n - 1) / (n * n - m * m))
            b = np.sqrt(((n - 1) ** 2 - m * m) / (4 * (n - 1) ** 2 - 1))
            P[(n, m)] = a * (ct * P[(n - 1, m)] - b * P[(n - 2, m)])
    Y = np.empty((len(ct), (N + 1) ** 2))
    for n in range(N + 1):
        for m in range(-n, n + 1):
            f = P[(n, abs(m))]
            Y[:, n * n + n + m] = (f if m == 0 else np.sqrt(2) * f *
                                   (np.cos(m * phi) if m > 0 else np.sin(-m * phi)))
    return Y


def mixing(N):
    """Q15 mixing matrix for order N: (K, K) int16, K = (N+1)^2 + 2. Gain is always 0."""
    U, _, Vt = np.linalg.svd(sh(N, nodes(N)))          # polar factor U.V^T: the nearest
    A = U @ Vt                                         # matrix with orthonormal columns
    K, J = (N + 1) ** 2 + 2, (N + 1) ** 2
    M = np.eye(K)
    M[:J, :J] = A
    # Q15, ties away from zero. The 1e-9 absorbs SVD rounding at exact ties (order 1 has
    # entries of exactly 1/2, i.e. 16383.5 counts), which differ by 1 ulp between LAPACKs.
    q = np.sign(M) * np.floor(np.abs(M) * 32767 + 0.5 + 1e-9)
    return A, q.astype(np.int16)


def committed(src):
    """{stem: int16 (K,K)} parsed out of an opus src/mapping_matrix*.c pair."""
    import re
    text = "".join(open(os.path.join(src, f)).read()
                   for f in ("mapping_matrix.c", "mapping_matrix_hoa.c"))
    out = {}
    for m in re.finditer(r"mapping_matrix_(\w+)_mixing = \{ (\d+), \d+, (-?\d+) \};\s*"
                         r"const opus_int16 \w+\[\d+\] = \{(.*?)\n\};", text, re.S):
        v = np.array([int(x) for x in re.findall(r"-?\d+", m.group(4))], np.int16)
        out[m.group(1)] = (v.reshape(int(m.group(2)), -1), int(m.group(3)))
    return out


STEMS = ("foa soa toa fourthoa fifthoa sixthoa seventhoa eighthoa ninthoa tenthoa "
         "eleventhoa twelfthoa thirteenthoa fourteenthoa").split()

ref = committed(sys.argv[1]) if len(sys.argv) > 1 else {}
print("order  nodes  cond(A)  SNR loss  min sep  Q15 orthog.  octets  vs. committed")
total = 0
for N in range(1, 15):
    A, q = mixing(N)
    J = (N + 1) ** 2
    x = nodes(N)
    d = np.degrees(np.arccos(np.clip(x @ x.T - 2 * np.eye(J), -1, 1))).min()
    # what the demixing step costs: ||A||_F ||A^-1||_F / J, 0 dB iff scaled-orthogonal
    loss = max(0.0, 20 * np.log10(np.linalg.norm(A) * np.linalg.norm(np.linalg.inv(A)) / J))
    Aq = q[:J, :J] / 32767.0                       # how orthogonal the table still is
    snr = -20 * np.log10(np.linalg.norm(Aq.T @ Aq - np.eye(J)) / np.sqrt(J))
    total += q.size * 2
    chk = ""
    if STEMS[N - 1] in ref:
        c, gain = ref[STEMS[N - 1]]
        chk = "match" if np.array_equal(c, q) and gain == 0 else "MISMATCH"
    print("%5d %6d %8.4f %9.2f %8.2f %11.1f %7d  %s" %
          (N, J, np.linalg.cond(A), loss, d, snr, q.size * 2, chk))
print("%d octets of .rodata for all 14 orders (%.1f KiB)" % (total, total / 1024))
```

</details>

### Where the numbers come from

The nodes are R. Womersley's maximum-determinant (Fekete) point sets, fetched per degree from `https://web.maths.unsw.edu.au/~rsw/Sphere/S2Pts/MD/md<degree:03d>.<npoints:05d>` (the [MaxDet index](https://web.maths.unsw.edu.au/~rsw/Sphere/MaxDet/); the full archive is also available as a 109 MB zip, hence the per-degree fetches). Degree N has exactly (N+1)^2 points, which is what makes the SH matrix square and the transform critically sampled.

From there it is four steps: evaluate the real orthonormal SH basis (ACN order, Condon-Shortley removed) at those nodes; take the polar factor `U.V^T` of the result, which is the nearest matrix with orthonormal columns; embed it in `K x K` alongside a 2x2 identity for the non-diegetic streams; quantize with `round(x * 32767)`, ties away from zero. Because the polar factor is invariant to a positive scale, the SH normalization constant washes out -- but the per-order `sqrt(2n+1)` does not, so the basis has to be N3D/orthonormal rather than SN3D. That is the only convention choice in the whole construction that is not forced.

`cond(A) = 1` and `SNR loss = 0.00 dB` at every order are the two properties the PR is really about: `||A||_F ||A^-1||_F / J` is exactly the factor by which per-sector coding noise is amplified when the decoder demixes it back to the SH domain, and it is 0 dB if and only if the mixing matrix is scaled-orthogonal.

### One caveat, for anyone reproducing this

The Q15 rounding carries a `+ 1e-9` guard. Order 1's polar factor has entries of exactly 1/2, i.e. 16383.5 counts -- a dead tie. Julia's LAPACK returns exactly 0.5 there; numpy's returns 0.5 - 1 ulp for three of the four entries, which floors to 16383 and gives a genuine one-count difference from the committed table. The guard sits about three orders of magnitude above the SVD noise and nine below any real quantization boundary, so it makes the tie rule deterministic across LAPACK implementations. Without it the script reproduces orders 2-14 exactly and order 1 off by one count in three cells.

Also note the `Q15 orthog.` column is not the same metric as the 46-65 dB the round-trip test prints: this is `||Aq^T Aq - I||_F / sqrt(J)`, while the test measures a round trip through both quantized matrices on a diffuse field. They are not in conflict.

jmvalin:
@chris-hld Can you have a look at this?

abhayap:
Thanks for looking at this Christoph. Your 4th/5th-order banks are close to optimal as sector banks — directivity 0.992 of the order-N maximum — so I want to be straight about what this does to them.

**The defect is order 2, which predates your tables.** `cond(A) = 11.74` against 1.3–3.3 either side, a `3050` Q8 gain field that costs ~2 of the 15 Q15 bits, and ~12 dB of SNR given away by the demixing step. Through libopus with a binaural render and PEAQ: −3.48 → −0.56 ODG at 64 kbit/s per channel.

**Why one construction rather than extending yours.** Fliege-Maier sets are cubature rules, not interpolation points, and past order 9 they stop working: `cond(A)` = 86 / 191 / 723 at orders 10 / 11 / 12, and the 121-point set is the first with a negative weight. Family 3 already accepts 227 channels, so something has to change up there regardless.

**In your framework.** The §4.1 condition for perfect reconstruction — a grid integrating degree 2N — can't be met at `L = (N+1)²` by any construction, yours included: (2N+1)² functions to be orthogonal, (N+1)² nodes to do it with. RFC8486 fixes the channel count, so this imposes `AᵀA = I` directly instead. What that buys is your §4.2 criterion, the one for when something alters the sector signals: energy preservation is exact at every order with `β_E = 1`, where the shipped tables miss it by 1.2–8.0 dB, and eq. (26) becomes the identity, so there is nothing for a decoder to apply.

**What it costs.** Orthogonalizing erases `cₙ` rather than softening it — ask for max-rE on these nodes and the weights come back flat. Side lobes −13.1 → −11.7 dB, directivity 0.992 → 0.969, and eq. (13) amplitude preservation, which the shipped orders 1–3 satisfy exactly, is not satisfied here. PEAQ measures |ΔODG| ≤ 0.06 at orders 1 and 3–7, so I'm not claiming these sound better; the case for touching 1–5 is uniformity, not quality.

**So if you'd rather 4 and 5 weren't touched, I'll split it.** Fixing order 2 and adding 6–12 needs neither. Say which shape you'd rather review.

PR #502, opened 2026-09-14
Correct "*X86_MAY_HAVE_AARCH64_NEON*" from meson.build

knutinh:
Correct apparent copy and paste error falsely indicating neon support on x86


jmvalin:
I'm not sure what the original code was intending on doing, but as far as I can tell, OPUS_ARM_MAY_HAVE_AARCH64_NEON_INTR isn't used anywhere in the code. @xnorpx Any thoughts?

PR #503, opened 2026-09-16
Reduce state-copy pressure in the hottest SILK mode

lpi:
Complexity-10 AArch64 NEON encoding with four delayed-decision states repeatedly copies one narrow state lane across the state rows. Preserve every fallback and state transition while scheduling that exact copy four rows at a time.

On an m4 mac processor this increases transcoding speed by 8%, on a aws c8g graviton, about 4%.


lpi:
I tested this, output is exactly the same across many different files.

lpi:
Tested head: `c97fd0b1d10fd59e5b1610e82fae8bfeae7801a9`.

The earlier directly qualified Graviton4 comparison against stock found the useful Candidate-3 result: production c10 encode improved thread CPU by **+4.0713%** and wall by **+4.0826%**, with c10 p99.9 **+4.7248%** and an optimized-symbol diagnostic of **+17.3194%**. Positive percentages mean faster/lower cost. The c5/c9 controls were approximately zero. This used 28 blocks and 22,400 frames per label per preset, with exact packet, final-range, and output identity. This remains the benchmark baseline.

